PR #22, opened 2020-12-02
Chore: Github actions

vikotar:
## Description
*Adding Puppeteer as a testing mechanism. So we are aware of when something breaks.*

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -1,0 +1,34 @@
+# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
+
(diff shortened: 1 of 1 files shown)


.github/workflows/node.js.yml
```diff
@@ -0,0 +1,34 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm ci
+    - run: npm run build --if-present
+#      move the secrets in github repo.
+      env:
```
vikotar, line 30: @abhishekgupta92 Could we put this into github repo secrets?
https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets

PR #23, opened 2020-12-02
Chore: Github actions

vikotar:
## Description
*Adding Puppeteer as a testing mechanism. So we are aware of when something breaks.*

P.S. The actions are failing. I have already worked on a few test cases in a new pr. It will start working when I will raise pr with that code.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -1,0 +1,33 @@
+# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
+
+name: Node.js CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/node.js.yml
```diff
@@ -0,0 +1,33 @@
...
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm ci
+    - run: npm run build --if-present
+#      move the secrets in github repo.
+      env:
+        REACT_APP_GOOGLE_CLIENT_ID:  ${{ secrets.REACT_APP_GOOGLE_CLIENT_ID }}
```
vikotar, line 31: @abhishekgupta92 Can we configure these secrets in the repo? https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets cc @pysaquib

PR #46, opened 2020-12-19
Feature/shareable exercise routes

tarun-dugar:


src/pages/CourseContent/index.js
```diff
@@ -37,25 +37,57 @@ function CourseContent(props) {
...
+            if (exerciseFromParams) {
+                defaultExercise = exerciseFromParams;
+                defaultExerciseIndex = data.exerciseList.findIndex((exercise) => {
+                    return exercise.id === exerciseIdFromParams;
+                });
+            }
+        }
+
+        // exerciseId not in params or exerciseId in params not there in exercise list (eg: invalid exercise id in url)
+        if (!defaultExercise) {
```
vikotar: I think we can put this info when we are declaring `defaultExercise` and `defaultExerciseIndex`. What do you think? one less check. 🙂

src/pages/CourseContent/index.js
```diff
@@ -37,25 +37,57 @@ function CourseContent(props) {
...
+
+        // set default exercise as per the exercise id present in the url
+        // TBD: Ideally, when navigating to a course content page, the first exercise should be pre-computed and the url should be course/:courseId/exercise/:exerciseId so we can always use the exerciseId from the params to set the default exercise.
+        if (exerciseIdFromParams) {
+            const exerciseFromParams = data.exerciseList.find((exercise) => {
+                return exercise.id === exerciseIdFromParams;
+            });
+            if (exerciseFromParams) {
+                defaultExercise = exerciseFromParams;
+                defaultExerciseIndex = data.exerciseList.findIndex((exercise) => {
```
vikotar: We might not need two loops to get the exercise from the `exerciseList` .  first, we get the index and then get it directly from the `exerciseList` by index access. i.e something like this. 
```
defaultExerciseIndex = data.exerciseList.findIndex((exercise) => {
                return exercise.id === exerciseIdFromParams;
            });
            defaultExercise = data.exerciseList[defaultExerciseIndex]
``` 
Let me know about your thoughts.

src/pages/CourseContent/index.js
```diff
@@ -37,25 +37,57 @@ function CourseContent(props) {
...
+    let exerciseIdFromParams = getExerciseIdFromUrl();
+    const firstExercise = get(data, 'exerciseList[0]');
+
+    // exercises loaded
     if (firstExercise) {
-      const selectedExerciseInfo = { exercise: firstExercise, index: 0 };
-      dispatch(courseActions.updateSelectedExercise(selectedExerciseInfo));
+
+        // set default exercise as per the exercise id present in the url
+        // TBD: Ideally, when navigating to a course content page, the first exercise should be pre-computed and the url should be course/:courseId/exercise/:exerciseId so we can always use the exerciseId from the params to set the default exercise.
```
vikotar, line 48: Nice suggestion. Let's create an Issue. and one of us can take it later. :)

src/pages/CourseContent/index.js
tarun-dugar: agreed, not sure why i did the other way round 😅

src/pages/CourseContent/index.js
tarun-dugar: agreed!

src/pages/CourseContent/index.js
vikotar: Cool. Man. I was the observer. It's easy for me to see that. I would have done the same I think. 😝

vikotar (approved):
Cool!! LGTM. thanks.  @tarun-dugar  
could you change the base branch to `dev`? I think we can merge it after that. 
Probably we should also change the default branch to `dev` so it picks `dev` as the base branch when we open a PR. cc @pysaquib

PR #57, opened 2021-01-02
Full name in  language field

komalahire:
- In the create classes form there is an option of language in that when the user is choosing the language for ex. English is there so only showing like this (en) but now in UI, it is showing the whole word in that field.

### DIFF
--- a/src/components/Class/ClassList/index.js
+++ b/src/components/Class/ClassList/index.js
@@ -11,12 +11,24 @@ function ClassList() {
 
   const { loading, data = [] } = useSelector(({ Class }) => Class.allClasses);
 
-  useEffect(() => {
-    dispatch(classActions.getClasses());
-  }, [dispatch]);
+  useEffect(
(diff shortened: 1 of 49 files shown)


src/components/Class/ClassList/index.js
```diff
@@ -11,12 +11,24 @@ function ClassList() {
 
   const { loading, data = [] } = useSelector(({ Class }) => Class.allClasses);
 
-  useEffect(() => {
-    dispatch(classActions.getClasses());
-  }, [dispatch]);
+  useEffect(
+    (e) => {
```
tarun-dugar: @komalahire can you remove 'e'. Also, did you not get any eslint errors when trying to make a git commit?

komalahire: Done,  yes I got error I removed also but i forget do again git add.

komalahire: Done

PR #61, opened 2021-01-19
Feat/stored last exercise

komalahire:
-  The last exercise cookies Saved in localStorage and show inside the card on the page bottom right. Continue where you left off.
Screenshot from 2021-01-19 10-35-50


src/components/App/ContinueExercise/index.js
```diff
@@ -0,0 +1,27 @@
+import React from "react";
+
+import "./styles.scss";
+
+let storedExercise = window.localStorage.getItem("last-exercise");
+const storedExerciseName = window.localStorage.getItem("exerciseName");
+
+function ContinueExercise() {
+  const [show, setShow] = React.useState(true);
```
tarun-dugar: add a condition: `if (!storedExercise) return null`;

komalahire: Done

PR #77, opened 2021-05-10
Edit class

tarun-dugar:


src/components/Class/ClassCard/index.js
```diff
@@ -69,12 +69,18 @@ function ClassCard(props) {
...
-            <button
-              className="delete-button"
-              onClick={() => deleteHandler(item.id)}
-            >
-              Delete
-            </button>
+            <div className="class-card-actions">
+              <i
+                className="class-card-action-icon fa fa-trash"
+                onClick={() => deleteHandler(item.id)}
```
komalahire: @tarun-dugar  Can we please use handleClickOpen instead of deleteHandler .  because whenever I am going to delete the class. without clicking the class is deleting.

tarun-dugar: nice catch! fixed it 👍

PR #78, opened 2021-05-10
Chat Rooms

Poonam-Singh-Bagh:


src/components/User/index.js
```diff
@@ -0,0 +1,91 @@
...
+    });
+  }, []);
+
+  const changeHandler = (e) => {
+    setEmail(e.target.value);
+  };
+
+  const submitHandler = () => {
+    const notify = () => {
+      toast.success("Added the class successfully", {
```
pysaquib: Please change the message to `Added ${user_email} to ${class room name}`

@Poonam-Singh-Bagh

pysaquib (changes requested):

PR #81, opened 2021-05-13
Managed create class api in component instead of redux

komalng:


src/components/Class/index.js
```diff
@@ -67,10 +65,40 @@ function Class() {
       payload[TIME_CONSTANT.CLASS_END_DATE] = `${moment(classEndTime).format(
```
tarun-dugar, line 65: @komalng can you delete the redux code that has the create class api as well?

komalng: https://github.com/navgurukul/bhanwari-devi/pull/81/commits/34a7b90788c3968197dbb480953e03159cce28a7

tarun-dugar (approved):

PR #86, opened 2021-05-22
Chat sdk

tarun-dugar:
Contains basic implementation of the meraki chat with the following features:

- send message
- delete message
- leave room
- reply to message
- text, interactive bot messages
- pagination


nilshah98:
Can someone share navgurkul credits to login and check
@pysaquib 

https://pr-86.dkchei85ij0cu.amplifyapp.com/

nilshah98:
Hi, @tarun-dugar was looking at the files and folder structures to get a top-down idea.
Have a question - what is the difference between message and messages under `mentor` ?
Note, I have not went through all the code, so this might be a dumb quesiton.

package.json
```diff
@@ -44,7 +46,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "lint-staged; ./node_modules/.bin/eslint . --ext .js"
+      "pre-commit": "lint-staged; ./node_modules/.bin/eslint src/**/*.js --ext .js"
```
nilshah98, line 52: Just a question -
Why was this edited, was `node_modules` also being checked ?

src/common/utils.js
```diff
@@ -0,0 +1,16 @@
+export const getIsMobile = () => {
```
nilshah98, line 1: Another question, any reference where this was taken from ?
And how does this work ?
_My assumption is looking and parsing the request headers ?_

src/components/common/Dropdown/index.js
```diff
@@ -0,0 +1,25 @@
...
+              onClick={option.onClick}
+            >
+              {option.label}
+            </button>
+          );
+        })}
+      </div>
+    );
+  }
+  return null;
```
nilshah98, line 24: Is returning `null` from a  component a good practice ?
Unaware about this.

src/pages/Mentor/ChatInput/index.js
```diff
@@ -0,0 +1,88 @@
...
+import "./styles.scss";
+
+export default ({
+  onNewMessage,
+  roomId,
+  replyMessage,
+  members,
+  activateReplyToMessageState,
+}) => {
+  const [value, onChange] = useState("");
```
nilshah98, line 13: By naming convention it should be - `value` and `setValue`, correct ?
This might be confusing

package.json
tarun-dugar: It's to explicitly specify what files to check for lint errors

src/common/utils.js
tarun-dugar: It checks against the user agent: https://stackoverflow.com/questions/11381673/detecting-a-mobile-browser

src/components/common/Dropdown/index.js
tarun-dugar: Yes, if we don't want to render anything then returning null is fine.

src/pages/Mentor/ChatInput/index.js
tarun-dugar: Not really, onChange is commonly used for form elements. Even native HTML elements have the `value` and `onChange` attributes.

tarun-dugar:
> Hi, @tarun-dugar was looking at the files and folder structures to get a top-down idea.
> Have a question - what is the difference between message and messages under `mentor` ?
> Note, I have not went through all the code, so this might be a dumb quesiton.

`Messages` contains logic for cases that pertain to the entire messages list as a whole. For example, showing the day divider in the chat. `Message` component only handles one message.

src/pages/Mentor/ChatInput/index.js
nilshah98: Oh, got it.
Never thought of it from this angle.

komalahire (approved):

PR #104, opened 2021-07-10
Create class exercise course

tarun-dugar:


pysaquib (approved):

src/components/Class/index.js
```diff
@@ -334,62 +364,102 @@ function Class({ classToEdit }) {
...
+                Maximum Enrollments
+                <span className="optional-field">(optional)</span>
+              </label>
               <input
                 className="input-field"
                 type="number"
-                name="max_enrolment"
-                id="max_enrolment"
-                min="1"
+                name={MAX_ENROLMENT}
```
nilshah98, line 457: This is an `input` tag and not `select`.
So, I think the name should just be string, "max_enrolment"

Any reasoning behind keeping it as a value ?

src/components/Class/index.js
```diff
@@ -39,27 +41,33 @@ function Class({ classToEdit }) {
...
-      ? moment.utc(end_time).format("kk:mm")
-      : moment().add(15, "minute").format("kk:mm"),
-    [LANG]: lang || "hi",
-    [TYPE]: type || "doubt_class",
-    [COURSE_ID]: course_id || "",
-    [PATHWAY_ID]: pathway_id || "",
-  });
+  const initialFormState = useMemo(() => {
+    return {
+      [TITLE]: title || "",
```
nilshah98, line 50: Just wanted to pick your brain about this.
I haven't encountered this syntax earlier, is it some pattern or new feature ?
ie - naming such as - [TITLE] = title, the square branches ( '[' ), do those signify something ?

Thanks.

PR #116, opened 2021-07-23
[FEAT] Added HiddenContent component

nilshah98:
**Features:**
- Added component `HiddenContent` to show and hide content
- Integrated `HiddenContent` with `ExerciseContent` component where `type == "solution"
- Preview -
Preview

**Edit:**
- Can change CSS to use `display : none` instead of using `visibility` if lots of white space is being left for the solution.

### DIFF
--- a/src/components/Course/Content/ExerciseContent/index.js
+++ b/src/components/Course/Content/ExerciseContent/index.js
@@ -7,6 +7,7 @@ import rehypeSanitize from "rehype-sanitize";
 import YouTube from "react-youtube";
 import get from "lodash/get";
 import { Prism as SyntaxHighlighter } from "react-syntax-highlighter";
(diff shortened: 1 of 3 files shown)


abhishekgupta92:
This is terrific.

src/components/Course/Content/ExerciseContent/index.js
```diff
@@ -113,6 +114,14 @@ const RenderContent = ({ data }) => {
       </code>
     );
   }
+  if (data.type === "solution") {
+    return (
+      <HiddenContent>
+        <ReactMarkdown components={components} children={data.value} />
```
tarun-dugar, line 120: @nilshah98 where are we using the `components` prop?

tarun-dugar (approved):

PR #131, opened 2021-08-04
1. Updating user's partner_id whenever they login/signup using a partner specific link

pysaquib:
1. Updating user's partner_id whenever they login/signup using a partner specific web app link
   - extract `referrer` from query string
   - call `PUT - /users/me` if referrer is present in query string

### DIFF
--- a/src/pages/Login/index.js
+++ b/src/pages/Login/index.js
@@ -1,13 +1,21 @@
-import React from "react";
+import React, { useState } from "react";
 import { useSelector, useDispatch } from "react-redux";
 import { Redirect } from "react-router";
 import GoogleLogin from "react-google-login";
+import axios from "axios";
+import { METHODS } from "../../services/api";
 import { actions as userActions } from "../../components/User/redux/action";
 import { PATHS } from "../../constant";
+import { getQueryVariable } from "../../common/utils";
 
 import "./styles.scss";
 
 function Login() {
+  const [queryString, setqueryString] = useState(null);
+
+  const updateQueryString = (value) => {
+    setqueryString(value);
+  };
   const dispatch = useDispatch();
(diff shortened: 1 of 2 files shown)


komalahire (approved):

src/pages/Login/index.js
```diff
@@ -22,9 +30,9 @@ function Login() {
       email: profile.getEmail(),
       idToken,
     };
-
     // let's send the data to our backend.
     dispatch(userActions.onUserSignin(googleData));
+    updateQueryString(getQueryVariable("referrer"));
```
nilshah98, line 35: There are two issues here -
1. You are doing the state update outside of the `useEffect` hook, this will cause a re-render each time.
2. Why aren't we using the single call we make in line 34, for sending the referrer name ? 
It calls - https://github.com/navgurukul/bhanwari-devi/blob/da55469ff742b93fab7f36eab908aa292fbe7536/src/components/User/redux/api.js#L12
Which in turn, makes a `POST` request to `${process.env.REACT_APP_MERAKI_URL}/users/auth/google`
Why can't we handle it in the same call itself ?

nilshah98 (changes requested):
There are two issues here -
1. You are doing the state update outside of the `useEffect` hook, this will cause a re-render each time.
2. Why aren't we using the single call we make in line 34, for sending the referrer name ? 
It calls - https://github.com/navgurukul/bhanwari-devi/blob/da55469ff742b93fab7f36eab908aa292fbe7536/src/components/User/redux/api.js#L12
Which in turn, makes a `POST` request to `${process.env.REACT_APP_MERAKI_URL}/users/auth/google`
Why can't we handle it in the same call itself ?

PR #133, opened 2021-08-05
Re-rendering the page after successful link creation

Poonam-Singh-Bagh:
Re-render the page after the successful link creation.


src/components/Dashboard /PartnerData/index.js
```diff
@@ -59,6 +59,7 @@ function PartnerDashboard() {
           position: toast.POSITION.BOTTOM_RIGHT,
           autoClose: 2500,
         });
+        setFlag(!flag);
```
pysaquib: @Poonam-Singh-Bagh Instead of setting a flag here, can we not set the state of the partner's meraki/app link (the one which just got created) and re-render the component based on this setState?

PR #135, opened 2021-08-16
Recurring classes

Poonam-Singh-Bagh:
All the 5 buttons ( enrollAll, drop out, deleteAll, editClass, Create class) are working.

### DIFF
--- a/src/components/Class/ClassCard/index.js
+++ b/src/components/Class/ClassCard/index.js
@@ -29,9 +29,8 @@ function ClassCard({ item, editClass, enroll, style, indicator }) {
     ta: "Tamil",
     doubt_class: "Doubt Class",
     workshop: "Workshop",
+    cohort: "Cohort",
   };
-
-  console.log("indicator", indicator);
 
   const handleClose = () => {
     setShowModel(false);
--- a/src/components/Class/ClassCard/index.js
+++ b/src/components/Class/ClassCard/index.js
@@ -140,6 +140,7 @@ function ClassCard({ item, editClass, enroll, style, indicator }) {
       dispatch(classActions.dropOutClass(Id));
     });
   };
+
   return (
     <div className="class-card ">
       <div className="class-details">
@@ -153,7 +154,7 @@ function ClassCard({ item, editClass, enroll, style, indicator }) {
           ) : null}
         </span>
         <h4>{item.title}</h4>
-        <p>Facilitator Name : {item.facilitator.name} </p>
+        <p>Facilitator : {item.facilitator.name} </p>
         <p>Language : {languageMap[item.lang]} </p>
         <p>Date:{moment(classStartTime).format("DD-MM-YYYY")} </p>
         {/* {item.email} */}
--- a/src/components/Class/ClassCard/index.js
+++ b/src/components/Class/ClassCard/index.js
@@ -12,7 +12,7 @@ import Modal from "../../common/Modal";
 
 toast.configure();
 
-function ClassCard({ item, editClass }) {
+function ClassCard({ item, editClass, enroll, style, indicator }) {
   const dispatch = useDispatch();
   const [enrollShowModel, setEnrollShowModel] = React.useState(false);
   const [unenrollShowModel, setunenrollShowModel] = React.useState(false);
@@ -30,6 +30,8 @@ function ClassCard({ item, editClass }) {
     doubt_class: "Doubt Class",
     workshop: "Workshop",
   };
+
+  console.log("indicator", indicator);
 
   const handleClose = () => {
     setShowModel(false);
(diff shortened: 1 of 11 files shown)


src/pages/Login/index.js
```diff
@@ -32,8 +33,11 @@ function Login() {
       idToken,
     };
     // let's send the data to our backend.
+    const referrer = getQueryVariable("referrer");
+    console.log(referrer);
```
tarun-dugar: remove console.log

src/pages/Login/index.js
```diff
@@ -24,6 +24,7 @@ function Login() {
   function onSignIn(googleUser) {
```
tarun-dugar, line 24: you have tested these changes, right?

src/components/Class/styles.scss
```diff
@@ -24,11 +24,15 @@
   .title {
     width: 100%;
     display: flex;
-    justify-content: center;
-    text-transform: uppercase;
+    // justify-content: center;
+    // text-transform: uppercase;
+    margin-left: 21%;
+    font-size: 200%;
```
tarun-dugar: Use px, not percentage. Also, why did we remove justify-content and text-transform? Please also test those components where these styles are being used.

src/components/Class/styles.scss
```diff
@@ -12,9 +12,9 @@
     width: 100%;
   }
 
-  label {
-    font-weight: bold;
-  }
+  // label {
+  //   font-weight: normal;
+  // }
```
tarun-dugar: why? this will affect all the labels in the app

src/components/Class/ClassList/index.js
```diff
@@ -21,20 +21,52 @@ function ClassList({ editClass, isShow }) {
...
+            {recurring_classes.map((item, index) => {
+              return (
+                <ClassCard
+                  item={item}
+                  key={index}
+                  index={index}
+                  editClass={editClass}
+                  enroll="Enroll to Cohort class"
+                  style="class-enroll-cohort"
+                  indicator="true"
```
tarun-dugar: true and false should be boolean values and not string.

src/components/Class/ClassCard/index.js
```diff
@@ -12,7 +12,7 @@ import Modal from "../../common/Modal";
```
tarun-dugar, line 12: please run the changes through Narendra before merging

src/components/Class/index.js
```diff
@@ -219,160 +258,246 @@ function Class({ classToEdit }) {
...
+                    onChange={(e) => {
+                      setFormField("workshop", TYPE);
+                    }}
+                    value={formFieldsState[TYPE]}
+                    id="type"
+                  />
+                  Workshop
+                </label>
+                <label htmlFor="type">
+                  <input
```
tarun-dugar, line 314: the id should be different for the three radio buttons

src/components/Class/index.js
```diff
@@ -219,160 +258,246 @@ function Class({ classToEdit }) {
...
-        {isEditMode ? "Update class" : "Create a class"}
+        {isEditMode ? "Update class" : "Create Cohort / Single Class"}
       </h2>
       <Form
         className="form"
         onSubmit={onFormSubmit}
         initialFieldsState={initialFormState}
       >
-        {({ formFieldsState, setFormField }) => {
+        {({ formFieldsState, setFormField, setFormFieldsState }) => {
```
tarun-dugar, line 282: why is setFormFieldsState required? We were using `setFormField` to set the state of a field before. Why is that not sufficient?

src/components/Class/index.js
```diff
@@ -178,6 +192,31 @@ function Class({ classToEdit }) {
     }
   };
 
+  const changeHandler = async (e, setField, field) => {
```
tarun-dugar, line 195: this handler is not needed, just use `setFormField`. If you want to set two fields, you can do:

setFormField(e.target.value, [TYPE]);
setFormField('', 'until')

Poonam-Singh-Bagh: This we need to use to handle occurrence and until. If a user is typing in the field until and the in occurrence then until field should get empty and if the user first types in occurrence and the until then occurrence field should get empty. The rest of the things will resolve and will let you know.

src/components/Class/styles.scss
Poonam-Singh-Bagh: I had removed text-transform because in design the title was in small later so I thought we need to keep it like this that's why.

src/components/Class/index.js
Poonam-Singh-Bagh: `setFormField` we have to give the key and value one at a time but we need to do `{ ...field, [e.target.name]: e.target.value, until: "" }` to make until empty when it's occurrence and vice versa. Hence, using setFormFieldsState as `setFormFieldsState({ ...field, [e.target.name]: e.target.value, until: "" })`

Poonam-Singh-Bagh: `setFormFieldsState({ ...formFieldsState, [e.target.name]: e.target.value, until: "" })`

tarun-dugar (dismissed):

PR #142, opened 2021-08-30
Fixed time issue in edit class

Poonam-Singh-Bagh:
Fixed time issue in edit class, can edit single class of cohort class and freeze cohort fields in edit class.


src/components/Class/index.js
```diff
@@ -92,8 +92,8 @@ function Class({ classToEdit }) {
   );
 
   const editClass = (payload) => {
-    if (classToEdit.recurring_id) {
-      payload = { ...payload, updateAll: true };
+    if (classToEdit.type === "cohort") {
+      delete payload.frequency;
```
pysaquib, line 96: @Poonam-Singh-Bagh Is frequency hard coded in the payload and that's why you're deleting it? Are you also deleting other cohort related fields like `until` or `occurrence` and `on_days` if not hardcoded?

Poonam-Singh-Bagh: Frequency is hard coded and as backend won't accept frequency if there is no updateAll so deleting frequency for now. No, not deleting cohort field they all are freezer in edit.

pysaquib (approved):

PR #145, opened 2021-09-05
redirect to the actual route user tried to visit after login

komalahire:
Redirect to the actual route user tried to visit after login


pysaquib:
@komalahire All good but https://pr-145.dkchei85ij0cu.amplifyapp.com/partners/355 doesn't work, so will any specific partner link

PR #154, opened 2021-09-16
Cohort class

Poonam-Singh-Bagh:
Users can able to delete, edit, enroll and drop all the classes of cohort.

1. Clicking on the delete, enroll, and drop button will take the user to a modal that will ask whether you want to do the function with all classes of cohort or just with a single class by clicking on the check box.
2. Clicking on the update class button will take the user to a modal that will ask whether the user wants to edit all cohort classes or a single class.


src/components/Class/ClassCard/index.js
```diff
@@ -12,12 +12,14 @@ import Modal from "../../common/Modal";
 
 toast.configure();
 
-function ClassCard({ item, editClass, enroll, style, indicator }) {
+function ClassCard({ item, editClass, enroll, style }) {
   const dispatch = useDispatch();
   const [enrollShowModel, setEnrollShowModel] = React.useState(false);
   const [unenrollShowModel, setunenrollShowModel] = React.useState(false);
   const [showModel, setShowModel] = React.useState(false);
```
pysaquib: @Poonam-Singh-Bagh Rename Model as Modal everywhere

src/components/Class/index.js
```diff
@@ -98,7 +98,9 @@ function Class({ classToEdit }) {
 
   const editClass = (payload) => {
     if (classToEdit.type === "cohort") {
-      delete payload.frequency;
+      if (indicator === false) {
+        delete payload.frequency;
```
pysaquib, line 104: @Poonam-Singh-Bagh Are you also removing other cohort related keys from the payload like `occurrence`, `on_days` or `until`

pysaquib (commented):
check each comment

src/components/Class/index.js
Poonam-Singh-Bagh: No. Because we are not sending these fields as payload when the user wants to edit only one class.

src/components/Class/ClassCard/index.js
Poonam-Singh-Bagh: Sure! Will rename it.

PR #164, opened 2021-09-25
Fix/disable update button

Poonam-Singh-Bagh:
 The user was able to update the class without even make any changes.

1. Compared the initial state and field state. If both of the states are the same then the update button would be disabled or else enable.

### DIFF
--- a/src/components/Class/index.js
+++ b/src/components/Class/index.js
@@ -304,6 +304,7 @@ function Class({ classToEdit, indicator }) {
         initialFieldsState={initialFormState}
       >
         {({ formFieldsState, setFormField, setFormFieldsState }) => {
+          const checkEquivalence = _.isEqual(initialFormState, formFieldsState);
           return (
             <>
               <label htmlFor="type">Select Class Type</label>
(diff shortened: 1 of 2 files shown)


src/components/Class/index.js
```diff
@@ -890,21 +873,33 @@ function Class({ classToEdit, indicator }) {
...
-                  <Loader />
-                ) : isEditMode ? (
-                  "UPDATE CLASS"
-                ) : (
-                  "Create Class"
-                )}
-              </button>
+              {_.isEqual(initialFormState, formFieldsState) ? (
+                <div class="cursor">
+                  <button
```
pysaquib: Don't render the button conditionally. Instead, set disable attribute conditionally. 
`  {const checkEquivalence = _.isEqual(initialFormState, formFieldsState)}
   <button
         disabled={checkEquivalence}
         className={checkEquivalence ? "pointer-disabled-class" : "pointer-enabled-class"}
`

pysaquib (changes requested):
Please resolve this

PR #167, opened 2021-10-12
Converting local time to IST before sending the payload in create and…

pysaquib:
… update class API

Local time is sent from the client. It can be from any timezone

1. Converting time to IST before sending the payload

### DIFF
--- a/src/components/Class/index.js
+++ b/src/components/Class/index.js
@@ -98,6 +98,8 @@ function Class({ classToEdit, indicator }) {
   );
 
   const editClass = (payload) => {
+    payload.start_time = convertToIST(payload.start_time);
+    payload.end_time = convertToIST(payload.end_time);
     if (classToEdit.type === "cohort") {
       if (indicator === false) {
         delete payload.frequency;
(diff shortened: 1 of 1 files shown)


src/components/Class/index.js
```diff
@@ -167,6 +169,15 @@ function Class({ classToEdit, indicator }) {
     });
   };
 
+  const convertToIST = (d) => {
+    const b = d.split(/\D+/);
+    const dateInObj = new Date(
+      Date.UTC(b[0], --b[1], b[2], b[3], b[4], b[5], b[6])
+    );
+    const utc = dateInObj.getTime() + dateInObj.getTimezoneOffset() * 60000;
```
saiky76, line 177: Assuming this file is in the front end code and running in browser,
does  dateInObj.getTimezoneOffset() return offset of this time to UTC? 
Is utc in millis here?

pysaquib: Yes, `getTimezoneOffset()` will return offset of whatever timezone the client is in wrt UTC. If the client is in New York timezone, it will return -300 since NY is -5:00 hours from UTC.

utc is in ms here

@saiky76

PR #179, opened 2021-10-27
Advanced course opportunity

komalahire:
Created advanced course opportunity page


src/pages/Opportunities/index.js
```diff
@@ -1,30 +1,16 @@
...
-      headers: {
-        accept: "application/json",
-        Authorization: Token,
-      },
-    }).then((res) => {
-      setUsers(res.data.user.partner_id);
-    });
-  }, []);
+  const url = window.location.href;
+  let partnerIdFromAndroid = url.split("partner_id=");
```
pysaquib: If you split it by `partner_id=` it doesn't guarantee that the element on position 1 will always be the `id`
For example, 
`http://merakilearn.org/login?partner_id=65` will work
BUT
`http://merakilearn.org/login?partner_id=65&referrer=breakthrough` will NOT
Here `partnerIdFromAndroid = 65&referrer=breakthrough`

src/pages/Opportunities/index.js
```diff
@@ -35,17 +21,16 @@ function Opportunities() {
         Authorization: user.data.token,
       },
     }).then((res) => {
-      // console.log(res, '===========')
       setPartner(res.data.partners);
     });
   }, []);
 
   let slug;
   for (const item of partner) {
```
pysaquib, line 29: Why not use _.filter?

src/pages/Opportunities/index.js
```diff
@@ -35,17 +21,16 @@ function Opportunities() {
...
-    if (item.id == userId) {
+    if (item.id == partnerId) {
```
pysaquib: Even if you're iterating through the partners array two if conditions can be combined into one
```
     if (item.id == partnerId || item.id == partnerIdFromAndroid[1]) {
            slug = item.slug
      }
```

pysaquib (changes requested):
Refer to comments

src/pages/Opportunities/index.js
```diff
@@ -1,7 +1,58 @@
...
+import { METHODS } from "../../services/api";
 
 function Opportunities() {
+  const [partner, setPartner] = useState([]);
+  const [users, setUsers] = useState();
+  const user = useSelector(({ User }) => User);
+  const userId = user.data.user.partner_id;
+
+  const Token =
+    "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpZCI6IjEzMTYiLCJlbWFpbCI6InBvb25hbUBuYXZndXJ1a3VsLm9yZyIsImlhdCI6MTYzNDU0ODc1OCwiZXhwIjoxNjY2MTA2MzU4fQ.PbvSCQbum3kigzwJs7BoSJPOFrqx6F70Rj6c-5AWOOo";
```
pysaquib: Why is this not in env variable?

pysaquib (changes requested):
Check comments

src/pages/Opportunities/index.js
komalahire: So the android team is sending partner id  in this format

https://www.merakilearn.org/opportunities?partner_id=value

nilshah98: @komalahire but in the future, they can send more parameters.
It is better to add an extra check as @pysaquib suggested.
It should not be a very tough task.

PR #193, opened 2021-12-06
Volunteer dashboard

komalahire:
Created volunteer dashboard feature.


src/components/VolunteerDashbord/VolunteerDetails/index.js
```diff
@@ -93,16 +96,26 @@ function VolunteerDashboard() {
     });
   }
 
+  function ratings() {
+    return cacheVolunteer.filter((el) => {
+      if (rating === "All") return true;
+      return rating == el.avg_rating;
+    });
```
jschanker: You can often succinctly combine conditionally executed `return` statements that each produce Boolean values (`true/false`):
```suggestion
    return cacheVolunteer.filter((el) => rating === "All" || rating == el.avg_rating);
```

src/components/VolunteerDashbord/IndividualVolunteers/index.js
```diff
@@ -141,6 +141,15 @@ function VolunteerOverview(props) {
           <tbody>
             {classes && classes.length > 0 ? (
               slicedClasses.map((item) => {
+                let ratingCount = 0;
+                item.ratings.map((item) => {
+                  if (item.rating) ratingCount += parseInt(item.rating);
+
+                  return ratingCount;
```
jschanker: From a purist standpoint, I wouldn't modify an existing Array in the callback function to map as the intended use of this functional programming construct is to create a new output Array from an existing input one by running each item through a function without side effects (e.g., `const doubledArr = arr.map(x => 2*x)`).  `forEach` would be more appropriate for an imperative approach when you're giving commands such as increment `ratingCount` by `parseInt(item.rating)`.  So, you could write something like
```
                item.ratings.forEach((item) => ratingCount += parseInt(item.rating) || 0);
```
Note the shorthand I used here; `parseInt` of a falsy value is either 0 (e.g., for `0`) or `NaN` (e.g., for `undefined` or `null`) so it'll use the (second) 0 which won't change the sum in these cases.

src/components/VolunteerDashbord/IndividualVolunteers/index.js
```diff
@@ -141,6 +141,15 @@ function VolunteerOverview(props) {
...
+                });
+                item.avg_rating = Math.ceil(
+                  item.ratings.length && ratingCount / item.ratings.length
```
jschanker: I'm not sure what a falsy value for `item.rating` means (no rating or 0 rating?), but assuming it means there was no rating, should we be counting that in the number of ratings when computing the average?  For example, if there are four 5-star ratings and one unavailable rating, should we effectively count it as 0 so that the average is 4 (20/5) instead of a 5-star average (20/4)?

src/components/VolunteerDashbord/IndividualVolunteers/index.js
```diff
@@ -141,6 +141,15 @@ function VolunteerOverview(props) {
...
+                );
```
jschanker: I don't know what the data looks like here so this could be off base, but based on what you wrote, something like this should work:
```suggestion
                const ratings = item.ratings.filter((item) => item.rating);
                item.avg_rating = ratings.length && Math.ceil(
                    ratings.reduce((ratingSum, item) =>
                    ratingSum + parseInt(item.rating), 0)/ratings.length
                    );
```
Here, by using `filter`, we can extract the actual `ratings` and store it in an Array `ratings`.  This means we won't have to use the trick above and can just divide the sum by the `length` of this `ratings` Array as it represents the list of only valid ratings.  Here, we used `reduce` for computing the sum, a common use case.

Please check this before committing though because I don't know what form the data is taking.

jschanker (commented):
Some suggestions

src/components/VolunteerDashbord/IndividualVolunteers/index.js
pysaquib: @komalahire This is what I was saying. @jschanker explained in much simpler way. Please look into this. You can increment count inside of 
`if (item.rating)` in line 146

src/components/VolunteerDashbord/IndividualVolunteers/index.js
pysaquib: @jschanker I don't think it is modifying the original array. Can you point out where it is doing so? Are you talking about `item` array in the callback of `slicedClasses.map( )`?

jschanker: @pysaquib Sorry, it's not being modified.  I copied a similar comment I made somewhere else where an Array was being modified but here I just meant that the variable `ratingCount` was being modified and the `return ratingCount` was not being used. (I.e., the code should work fine without the `return` statement in this `map`.)

src/components/VolunteerDashbord/IndividualVolunteers/index.js
pysaquib: This is in fact correct suggestion. It also resolves the average_rating problem we discussed above. But this is for average_rating per class. We still need to resolve the average_rating per volunteer. This is the structure of ratings key
`       "ratings": [
          {
            "rating": "5"
          },
          {
            "rating": null
          },
          {
            "rating": "3"
          }
        ]`
@komalahire

pysaquib: @jschanker Instead of `const avg_rating` we need to do `item.avg_rating`

src/components/VolunteerDashbord/IndividualVolunteers/index.js
pysaquib: That's right. @komalahire you can remove the return statement from map. Also, for avoiding the lint error, you can use `forEach` instead of `map`

src/components/VolunteerDashbord/IndividualVolunteers/index.js
```diff
@@ -0,0 +1,224 @@
...
+  useEffect(() => {
+    const data = volunteerClassData.classes.filter((searchValue) => {
+      if (searchTerm == "") {
+        return searchValue;
+      } else if (
+        searchValue.title.toLowerCase().includes(searchTerm.toLowerCase())
+      ) {
+        return searchValue;
+      }
+    });
```
jschanker, line 56: What you have should work, but generally, the callback function to `filter` should always **explicitly** return a value and that value should always be a `Boolean` with `true` meaning "use" and `false` meaning "do **not** use" for the resulting filtered `Array`.

Currently, you're returning a truthy value (object `searchValue`) in the cases that there was no entered `searchTerm` or the `searchTerm` can be found in the title of `searchValue`.  You're not explicitly returning anything otherwise so in that case you're implicitly returning the falsy `undefined`.  Instead, I'd suggest having the callback just answer the question should I keep this? by answering yes, exactly when `searchTerm == "" || searchValue.title.toLowerCase().includes(searchTerm.toLowerCase()))`.  But actually the first part is unnecessary since every string contains the (lowercase version) of the empty string, so my suggestion would be to just use the following:

```suggestion
    const data = volunteerClassData.classes.filter((searchValue) =>
        searchValue.title.toLowerCase().includes(searchTerm.toLowerCase()));
```

src/components/VolunteerDashbord/IndividualVolunteers/index.js
```diff
@@ -0,0 +1,224 @@
...
+    setPageNumber(selected);
+  };
+
+  const firstIndex = (pageNumber + 1) * limit;
+  const lastIndex = volunteerClassData.classes.length;
+
+  const lang = { en: "English", hi: "Hindi" };
+  let language = "";
+  new Set(volunteerClassData.classes.map((item) => item.lang)).forEach(
+    (item) => (language = language + lang[item] + ", ")
```
jschanker, line 37: Nit: Does it matter that there will be a trailing comma at the end of the list of languages (assuming at least one is used)?

src/components/VolunteerDashbord/IndividualVolunteers/index.js
```diff
@@ -0,0 +1,224 @@
...
+  const volunteerClassData = props.location.state.pass;
+  const pageCount = Math.ceil(volunteerClassData.classes.length / limit);
+  const changePage = ({ selected }) => {
+    setPageNumber(selected);
+  };
+
+  const firstIndex = (pageNumber + 1) * limit;
+  const lastIndex = volunteerClassData.classes.length;
+
+  const lang = { en: "English", hi: "Hindi" };
```
jschanker, line 34: Can we just move the `languageMap` definition up here, remove the `lang` declaration and then just replace the occurrences of `lang` in `item.lang` and `lang[item]` with `langMap`?  Or even better could the `languageMap` be moved out from this file and to a constants file so it could be used across the repository?  This would be helpful if classes were added in different languages in the future.

src/components/VolunteerDashbord/IndividualVolunteers/index.js
```diff
@@ -0,0 +1,224 @@
...
+                      {moment.utc(item.end_time).format("DD-MM-YYYY")}
+                    </td>
+                    <td data-column="Class Time">
+                      {moment
+                        .utc(item.start_time)
+                        .add(330, "minute")
+                        .format("kk:mm")}
+                    </td>
+                    <td data-column="Enrollments">
+                      {item.max_enrollment ? item.max_enrollment : "NA"}
```
jschanker, line 175:
```suggestion
                      {item.max_enrollment || "NA"}
```

src/components/VolunteerDashbord/IndividualVolunteers/styles.scss
```diff
@@ -0,0 +1,143 @@
...
+  box-shadow: 0px 0px 0px 0px;
+}
+
+.sort-class {
+  background: transparent;
+  border: none;
+}
+
+@media only screen and (max-width: 760px),
+  (min-device-width: 768px) and (max-device-width: 1024px) {
```
jschanker: The CSS feature `device-width` is deprecated now. (See: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/device-width and https://stackoverflow.com/a/18500871)

src/components/VolunteerDashbord/IndividualVolunteers/index.js
```diff
@@ -0,0 +1,224 @@
...
+          </div>
+          <div> / {volunteerClassData.name}</div>
+        </div>
+        <div className="volunteer-details">
+          <p className="volunteer-name">{volunteerClassData.name}</p>
+          <p>Python: {language.slice(0, -2)}</p>
+          <p>Total Classes: {volunteerClassData.classes.length}</p>
+          <p>
+            Associated Partner:{" "}
+            {volunteerClassData.partner ? volunteerClassData.partner : "NA"}
```
jschanker, line 100:
```suggestion
            {volunteerClassData.partner || "NA"}
```

src/components/VolunteerDashbord/IndividualVolunteers/index.js
jschanker: @pysaquib Yeah, I had used `avg_rating` for some quick console testing and had initially forgotten to change it back.  I updated my answer about an hour later, but I don't think notifications are sent for updates.  Anyway, I think the edited answer should be good.

src/components/VolunteerDashbord/IndividualVolunteers/index.js
Poonam-Singh-Bagh: language is a string and using a comma after each element to get a string like `"English, Hindi,"` so that we can display it as `English, Hindi`

jschanker: @Poonam-Singh-Bagh Thanks, yes, I see now you addressed the extra comma and space here by making a copy of the string with the last two characters removed: https://github.com/navgurukul/bhanwari-devi/blob/1597c7fa81ba906adbc057ac9b4e8bf65c42b4ae/src/components/VolunteerDashbord/IndividualVolunteers/index.js#L96

src/components/VolunteerDashbord/IndividualVolunteers/styles.scss
Poonam-Singh-Bagh: Okay, if we can't use it now, what can we use instead?

pysaquib: @Poonam-Singh-Bagh Why are we using `device-width` ? For responsiveness in device orientation? From what I understand, `device-width` takes your device screen size into account. If you want to target the viewport you can simply use `max-width` and `min-width` (or `max-height` and `min-height`)

Poonam-Singh-Bagh: I don't know Komal has used device-width in Partners' Dashboard and elsewhere we are using `@media screen and (max-width: 768px)` . I'm replacing `device-width` with `@media screen and (max-width: 768px)`

jschanker: @Poonam-Singh-Bagh As @pysaquib mentioned, `max-width`, `min-width`, `max-height`, and `min-height` are generally what we care about when making a responsive site since they are sensitive to changes in browser dimensions, allowing for changes to the layout to be applied on the fly during resizing, while `device-width` and `device-height` are not.  If the point of using `device-width` was to target say iPads, then a replacement could be to use a combination of other media features that might distinguish them (e.g., [hover](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/hover) of `none`, [resolution](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/resolution) of `192dpi`) but this almost certainly won't be 100% accurate as these features could be shared by non-iPads and are subject to changes.  However, this is also why `device-width` wasn't too useful for distinguishing devices, as dimensions of screens are not necessarily specific to say iPads and newer versions of these devices (e.g., iPad Pros) can wind up having different dimensions.

PR #194, opened 2021-12-06
Changed table component data from row to col to display table

Poonam-Singh-Bagh:
The table wasn't getting displayed properly

1. Changed the list data from row to column
2. Changed the place of <tr> and <td>


src/components/Course/Content/ExerciseContent/index.js
```diff
@@ -114,6 +114,28 @@ const RenderContent = ({ data }) => {
...
+      for (const j in dataInRow) {
+        if (j == 3) {
+          break;
+        } else {
+          temArray.push(dataInRow[j]);
+        }
+      }
+      dataInCol.push(temArray);
+      dataInRow.splice(0, 3);
+    }
```
jschanker: This could be done more succinctly as follows:
```suggestion
    const dataInCol = allData[0].map((_, i) => allData.map((_, j) => allData[j][i]));
```
This is basically what you came up with, but in lines 120-123, you essentially created a flattened version of this Array by pushing one element at a time and then undid the flattening in lines 127-138.

src/components/Course/Content/ExerciseContent/index.js
```diff
@@ -114,6 +114,28 @@ const RenderContent = ({ data }) => {
...
+    const dataInCol = [];
+    for (const i in allData[0]) {
+      const temArray = [];
+      for (const j in dataInRow) {
+        if (j == 3) {
+          break;
+        } else {
+          temArray.push(dataInRow[j]);
+        }
+      }
```
jschanker: This doesn't matter if you use the suggestion, but note that when iterating over the indices of an Array `dataInRow`, it's usually clearer to write this as:
`for (let j = 0; j < dataInRow.length; ++j)` 
(You probably already know this, but just in case you don't, this initializes `j` to `0`, checks that `j` is less than `dataRow.length`, if so proceeds to the body of the loop, then updates by adding `1` to `j` (`++j` is shorthand for `j = j + 1`), checks that this new value of `j` is less than `dataRow.length` and if so proceeds to the body of the loop, updates `j` by adding `1` to get `2`, etc.)
However, in this case, using this form has the additional benefit of allowing you to making your code shorter since you only want to continue while `j < 3`.  So the entire loop would become:
```
for (let j = 0; j < 3; ++j) {
  temArray.push(dataInRow[j]);
}
```
In general, you want to try to avoid using `break` in a loop if you can because it can make your code more difficult to understand.

src/components/Course/Content/ExerciseContent/index.js
```diff
@@ -114,6 +114,28 @@ const RenderContent = ({ data }) => {
...
+    for (const i in allData[0]) {
+      for (const j in allData) {
+        dataInRow.push(allData[j][i]);
+      }
+    }
+    const dataInCol = [];
+    for (const i in allData[0]) {
+      const temArray = [];
+      for (const j in dataInRow) {
+        if (j == 3) {
```
jschanker: Change is constant in software development, so you don't want to assume that the table will always have `3` columns.  Instead, you can get the number of columns in general by using `allData.length` (or `data.value.length`).

jschanker (commented):
Hope these are helpful suggestions!

PR #199, opened 2021-12-14
Displaying only 4 pathways

Poonam-Singh-Bagh:
Gave if condition 
if pathway is not Residential Programmes Info-Track then render it

### DIFF
--- a/src/components/Class/index.js
+++ b/src/components/Class/index.js
@@ -381,22 +381,26 @@ function Class({ classToEdit, indicator }) {
               <label htmlFor="pathway">Pathway</label>
               <span>
                 {pathways.map((item) => {
-                  return (
-                    <label htmlFor="pathway" className="radio-pointer">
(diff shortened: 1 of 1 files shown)


src/components/Class/index.js
```diff
@@ -381,22 +381,26 @@ function Class({ classToEdit, indicator }) {
...
-                        value={item.id}
-                        onChange={(e) => {
-                          setPathwayId(e.target.value);
-                        }}
-                        checked={pathwayId === `${item.id}` ? "checked" : false}
-                      />
-                      {item.name}
-                    </label>
-                  );
+                  if (item.name !== "Residential Programmes Info-Track") {
```
pysaquib: Check the if condition with `item.code`. It is more reliable than name.

pysaquib (changes requested):
Reviewed

PR #224, opened 2022-01-19
Fix/select radio

Poonam-Singh-Bagh:


src/components/Class/index.js
```diff
@@ -337,14 +342,22 @@ function Class({ classToEdit, indicator }) {
...
                       formFieldsState.type === "cohort" ? "checked" : false
                     }
+                    // disabled={formFieldsState[TYPE] === "cohort" ? true : false}
+                    disabled={
+                      isEditMode
+                        ? formFieldsState[TYPE] === "cohort"
+                          ? false
+                          : true
+                        : false
+                    }
```
jschanker, line 356: In order for this expression to evaluate to `true`, `isEditMode` needs to be `true` and `formFieldsState[TYPE] === "cohort"` needs to be `false` so you can write this as follows:
```suggestion
                    disabled={isEditMode && formFieldsState[TYPE] !== "cohort"}
```

src/components/Class/index.js
```diff
@@ -353,21 +366,31 @@ function Class({ classToEdit, indicator }) {
...
                     checked={
                       formFieldsState.type === "doubt_class" ? "checked" : false
                     }
+                    disabled={
+                      isEditMode
+                        ? formFieldsState[TYPE] === "doubt_class"
+                          ? false
+                          : true
+                        : false
+                    }
```
jschanker, line 379: Similar comment here:
```suggestion
                    disabled={isEditMode && formFieldsState[TYPE] !== "doubt_class"}
```
But also lines 336-359 and lines 360-383 are almost identical so you may want to consider creating a new Component to avoid this duplication.

komalahire (approved):

PR #227, opened 2022-01-21
Fix/select radio

Poonam-Singh-Bagh:
@jschanker This one

### DIFF
--- a/src/components/Class/index.js
+++ b/src/components/Class/index.js
@@ -345,13 +345,6 @@ function Class({ classToEdit, indicator }) {
                     checked={
                       formFieldsState.type === "cohort" ? "checked" : false
                     }
-                    // disabled={
-                    //   isEditMode
-                    //     ? formFieldsState[TYPE] === "cohort"
-                    //       ? false
-                    //       : true
-                    //     : false
-                    // }
                     disabled={isEditMode && formFieldsState[TYPE] !== "cohort"}
                   />
                   Cohort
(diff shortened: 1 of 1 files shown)


src/components/Class/index.js
```diff
@@ -371,11 +363,7 @@ function Class({ classToEdit, indicator }) {
                       formFieldsState.type === "doubt_class" ? "checked" : false
                     }
                     disabled={
-                      isEditMode
-                        ? formFieldsState[TYPE] === "doubt_class"
-                          ? false
-                          : true
-                        : false
+                      isEditMode && formFieldsState[TYPE] !== "doubt_class"
```
jschanker, line 366:
```suggestion
                    disabled={isEditMode && formFieldsState[TYPE] !== "doubt_class"}
```

jschanker (commented):
(Corrected below to account for closing `}` below)

src/components/Class/index.js
```diff
@@ -371,11 +363,7 @@ function Class({ classToEdit, indicator }) {
...
                     }
```
jschanker, line 367:
```suggestion
                    disabled={isEditMode && formFieldsState[TYPE] !== "doubt_class"}
```

jschanker (commented):
@Poonam-Singh-Bagh I was able to suggest a change for lines `365` to `367` because my suggestion removed whitespace. I can't for line `348` because my suggestion is exactly the same as what you have now. I also opened a separate pull request #228, where I made the changes myself, leaving in your `console.log` statement from line `334`, so you could use whichever is most convenient.

PR #229, opened 2022-01-22
Opportunity flow

komalahire:
**Brief description of the admission flow**

1. Created a component RedirectComponet which is getting used as a landing page for android.
2. After RedirectComponent if a user is logged in on the android side, will get redirected to the Admission where the user can fill out a form and can give an admission test.
3. After RedirectComponent if a user is not logged in on the android side, will get redirected to the Login page and after logged in will redirect to admission.

Use @Poonam-Singh-Bagh   @jschanker


src/pages/Navgurukul/merakiAdmission/index.js
```diff
@@ -0,0 +1,211 @@
...
+
+  const generateTestLink = async () => {
+    try {
+      const mobile = `0${userDetails.mobileNumber}`;
+      const dataURL = `${process.env.REACT_APP_CHANAKYA_BASE_URL}helpline/register_exotel_call`;
+      const response = await axios.get(dataURL, {
+        params: {
+          ngCallType: "getEnrolmentKey",
+          From: mobile,
+          partner_id: partnerId ? partnerId : null,
```
jschanker:
```suggestion
          partner_id: partnerId || null,
```

src/pages/Navgurukul/merakiAdmission/index.js
```diff
@@ -0,0 +1,211 @@
...
+        mobileNumber: userDetails.mobileNumber,
+      };
+      const queryString = Object.keys(params)
+        .map((filter) => {
+          if (params[filter]) {
+            return `${filter}=${params[filter]}`;
+          }
+          return null;
+        })
+        .filter((item) => item)
```
jschanker: Reversing the `map` and `filter` here removes the need for the `if` (and `return null`).
```suggestion
      const queryString = Object.keys(params)
          .filter((key) => params[key])
          .map((key) => `${key}=${params[key]}`);
```

src/pages/Navgurukul/styles.scss
```diff
@@ -0,0 +1,40 @@
...
+  align-self: center;
+  width: 290px;
+}
+.get-meraki-link a {
+  color: white;
+  text-decoration: none;
+}
+
+@media only screen and (max-width: 760px),
+  (min-device-width: 768px) and (max-device-width: 1024px) {
```
jschanker: `min-device-width` and `max-device-width` are deprecated.  See earlier discussion here: https://github.com/navgurukul/bhanwari-devi/pull/193#discussion_r766278657
```suggestion
@media only screen and (max-width: 768px) {
```

src/pages/Navgurukul/navgurukulLogin/index.js
```diff
@@ -0,0 +1,54 @@
+import React, { useState } from "react";
+import "./styles.scss";
+import GoogleLogin from "react-google-login";
+import { Redirect } from "react-router";
+import { useSelector, useDispatch } from "react-redux";
+import { actions as userActions } from "../../../components/User/redux/action";
+
+function NavgurukulLogin() {
```
jschanker: Much of this is duplicated on the [Login page](https://github.com/navgurukul/bhanwari-devi/blob/main/src/pages/Login/index.js).  Can the common logic/UI be factored into a `Login` component that can then be imported for both pages (and any future Login page you might want to create in the future)?

src/routing/index.js
```diff
@@ -35,6 +39,14 @@ const Routing = () => {
       <Route exact path={PATHS.AFE} component={AFEpage} />
       <Route exact path={PATHS.PROFILE} component={Profile} />
```
jschanker: Not related to this pull request, but should this be a private route?

src/pages/Navgurukul/merakiAdmission/index.js
```diff
@@ -0,0 +1,211 @@
...
+    try {
+      const mobile = `0${userDetails.mobileNumber}`;
+      const dataURL = `${process.env.REACT_APP_CHANAKYA_BASE_URL}helpline/register_exotel_call`;
+      const response = await axios.get(dataURL, {
+        params: {
+          ngCallType: "getEnrolmentKey",
+          From: mobile,
+          partner_id: partnerId ? partnerId : null,
+        },
+      });
```
jschanker: API call?

src/pages/Navgurukul/merakiAdmission/index.js
```diff
@@ -0,0 +1,211 @@
...
+                  name="lastName"
+                  onChange={changeHandler}
+                />
+              </div>
+            </div>
+            <div className="form-row">
+              <div className="input-field-test">
+                <span>Mobile Number</span>
+                <input
+                  type="text"
```
jschanker: You can use the `tel` input type with pattern validation here to require 10-digit phone numbers (See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/tel#pattern_validation) as follows.  However, the pattern for the input will not be enforced here as the inputs are not inside a `<form>` element.  

It may be useful to create a `PhoneNumberInput` component to remove hyphens and parentheses, look up entered phone numbers to make sure they're valid, etc.

```suggestion
                  type="tel"
                  pattern="^[0-9]{10}$"
```

src/pages/Navgurukul/merakiAdmission/index.js
```diff
@@ -0,0 +1,211 @@
...
+            <div className="input-field-test">
+              <span>Mobile Number</span>
+              <input
+                type="text"
+                placeholder="9874500000"
+                onChange={(e) => {
+                  setMobile(e.target.value);
+                }}
+                value={mobile}
+              />
```
jschanker:
```suggestion
              <label>
                Mobile Number:
                <input
                  type="tel"
                  pattern="[0-9]{10}"
                  placeholder="9874500000"
                  onChange={(e) => {
                    setMobile(e.target.value);
                  }}
                  value={mobile}
                />
              </label>
```

src/pages/Navgurukul/merakiAdmission/index.js
```diff
@@ -0,0 +1,211 @@
...
+    ) {
+      toast.error(
+        "To attempt the test, it is compulsory to enter your First Name, Last Name and Mobile Number. Middle Name is optional, you can choose not to enter.",
+        {
+          position: toast.POSITION.BOTTOM_RIGHT,
+        }
+      );
+      return;
+    }
+    if (userDetails.mobileNumber.toString().length !== 10) {
```
jschanker: You can use a [regular expression test](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test) to make sure that the string only consists of (exactly 10) digits:
```suggestion
    if (!/[0-9]{10}/.test(userDetails.mobileNumber.toString())) {
```

src/components/common/RedirectComponent/index.js
```diff
@@ -0,0 +1,23 @@
...
+  const token = getQueryVariable("token");
+  const redirect = getQueryVariable("redirectUrl");
+
+  localStorage.setItem("Token", token);
+
+  if (uri.includes("token")) {
+    return <Redirect to={redirect} />;
+  } else {
+    return <Redirect to="/admission-login" />;
+  }
```
jschanker:
```suggestion
  return <Redirect to={uri.includes("token") ? redirect : "/admission-login"} />;
```

src/pages/Navgurukul/styles.scss
Poonam-Singh-Bagh: Thank you @jschanker to remind me again. Will keep this in mind.

src/pages/Navgurukul/merakiAdmission/index.js
Poonam-Singh-Bagh: Yes.

src/routing/index.js
Poonam-Singh-Bagh: No. They have told us to keep AFE open route so that anyone can reach this page. And Profile should be a private route, making the changes. Thanks to drag the attention here.

src/pages/Navgurukul/merakiAdmission/index.js
Poonam-Singh-Bagh: If I do like this with `[0-9]{10}` then it won't throw an error when the length is greater than 10 . So I'm doing like this `^[0-9]{10}$`

src/pages/Navgurukul/navgurukulLogin/index.js
Poonam-Singh-Bagh: Sure. Thanks

src/pages/Navgurukul/merakiAdmission/index.js
Poonam-Singh-Bagh: @jschanker even after using patter with a limit of 10, we can use more than 10, so why do we need to use pattern? 
Also `type='tel'` can take alphabets, so why need to use it, we can use `type='number'` instead?

jschanker: @Poonam-Singh-Bagh Using `type='tel'` gives the browser (and search engines) better semantic information about the form element.  If you use `type = 'number'`, then you may have up/down arrows on the right of the form field, which could be confusing to the user as you don't usually enter telephone numbers that way.  As for `pattern`, this again provides more information to the browser.  If you have the following HTML, then modern browsers will often prevent the form from being submitted when the user enters a number that isn't exactly 10 digits with a validation message appearing under the text box, saying something like, "Please match the requested format."  This message will pop up only *after* you click the submit button and will not occur if the input appears outside a form.  However, even in cases where it has no functional purpose for a browser, search engines can still make use of this extra information.  Also, I updated the regex with the `^` and `$`, as you had mentioned, to make sure it wasn't matching a proper substring of an invalid input.  Sorry about my mistake before.
```
<form>
  <input type="tel" pattern="^[0-9]{10}$" />
  <button type="submit">Submit</button>
</form>
```

jschanker:
@Poonam-Singh-Bagh Just a note that one of my clarifications appears in the collapsed 12 review comments so you may not be able to get it directly when clicking on the link.

src/pages/Navgurukul/index.js
```diff
@@ -0,0 +1,46 @@
...
+
+  const getMerakiLink = () => {
+    axios({
+      method: METHODS.PUT,
+      url: `${process.env.REACT_APP_MERAKI_URL}/partners/${partnerId}/merakiLink`,
+      headers: {
+        accept: "application/json",
+        platform: "android",
+      },
+    }).then((res) => {
```
jschanker, line 19: API call?

src/pages/Navgurukul/merakiAdmission/index.js
```diff
@@ -0,0 +1,211 @@
...
+  let userToken = localStorage.getItem("Token");
+  useEffect(() => {
+    axios({
+      method: METHODS.GET,
+      url: `${process.env.REACT_APP_MERAKI_URL}/users/me`,
+      headers: {
+        accept: "application/json",
+        Authorization: userToken,
+      },
+    }).then((res) => {
```
jschanker: Should this be done with an API call that you could define [here](https://github.com/navgurukul/bhanwari-devi/blob/main/src/components/User/redux/api.js)?  You would also be able to make the API call from the Profile page as well: https://github.com/navgurukul/bhanwari-devi/blob/8dc95154ddccc11eb54f1a7a220b2df907bb253d/src/pages/Profile/index.js#L15-L22

src/pages/Navgurukul/merakiAdmission/index.js
```diff
@@ -0,0 +1,211 @@
...
+    }
+    await isDuplicate();
+  };
+
+  const isDuplicate = () => {
+    const { firstName, middleName, lastName, mobileNumber } = userDetails;
+    axios
+      .get(`${process.env.REACT_APP_CHANAKYA_BASE_URL}/check_duplicate`, {
+        params: {
+          Name: firstName.concat(" ", middleName, lastName),
```
jschanker: Although not a performance issue here, using `+` is preferred for concatenating strings (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/concat#performance).

src/pages/Navgurukul/merakiAdmission/index.js
```diff
@@ -0,0 +1,211 @@
...
+          Number: mobileNumber,
+        },
+      })
```
jschanker: API call?

src/pages/Navgurukul/merakiAdmission/index.js
```diff
@@ -0,0 +1,211 @@
...
+      .then(async (data) => {
+        const response = data.data.data;
+        const stage = response.pendingInterviewStage;
+        const url = `https://admissions.navgurukul.org/check_duplicate/Name=${firstName}${middleName}${lastName}&Number=${mobileNumber}&Stage=${stage}`;
```
jschanker: Could we create a constant like `ADMISSIONS_URL` in a config file and assign it to `https://admissions.navgurukul.org/` so we could import it instead of hard-coding it?  This way, if the URL were to ever change, it could be updated in one central place without having to find and replace all of the occurrences of it in source code.

src/pages/Navgurukul/merakiAdmission/index.js
```diff
@@ -0,0 +1,211 @@
...
+          <div className="input-form">
+            <div className="form-row">
+              <div className="input-field-test">
+                <span>First Name</span>
+                <input
+                  type="text"
+                  placeholder="Abhi..."
+                  value={userDetails.firstName}
+                  name="firstName"
+                  onChange={changeHandler}
```
jschanker, line 143: It may be good to add the [`required`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required) attribute here and on the Last Name, and Mobile Number text inputs.
```suggestion
                  required
                  onChange={changeHandler}
```

src/pages/Navgurukul/merakiAdmission/index.js
```diff
@@ -0,0 +1,211 @@
...
+                />
+              </div>
+              <div className="input-field-test">
+                <span>Last Name</span>
+                <input
+                  type="text"
+                  placeholder="Garg..."
+                  value={userDetails.lastName}
+                  name="lastName"
+                  onChange={changeHandler}
```
jschanker, line 166:
```suggestion
                  required
                  onChange={changeHandler}
```

src/pages/Navgurukul/merakiAdmission/index.js
```diff
@@ -0,0 +1,211 @@
...
+            </div>
+            <div className="form-row">
+              <div className="input-field-test">
+                <span>Mobile Number</span>
+                <input
+                  type="text"
+                  placeholder="9874500000"
+                  value={userDetails.mobileNumber}
+                  name="mobileNumber"
+                  onChange={changeHandler}
```
jschanker, line 181:
```suggestion
                  required
                  onChange={changeHandler}
```

src/pages/Navgurukul/merakiAdmission/index.js
```diff
@@ -0,0 +1,211 @@
...
+                  setMobile(e.target.value);
+                }}
+                value={mobile}
+              />
+            </div>
+          </div>
+          <button className="test-btn">
+            <a
+              className="result-btn"
+              href={`https://admissions.navgurukul.org/status/${mobile}`}
```
jschanker: Same comment here about referring to constant for https://admissions.navgurukul.org

src/routing/index.js
```diff
@@ -35,6 +39,14 @@ const Routing = () => {
...
       <Route exact path={PATHS.OPPORTUNITIES} component={Opportunities} />
+      <Route
+        exact
+        path={PATHS.NAVGURUKUL_INTRODUCE}
+        component={NavgurukulIntroduce}
+      />
+      <Route exact path={PATHS.ADMISSION_LOGIN} component={NavgurukulLogin} />
+      <Route exact path={PATHS.ADMISSION} component={Admission} />
```
jschanker: @Poonam-Singh-Bagh Should `ADMISSION` be a private route?  Otherwise the user is able to navigate directly to `/admission` without the presence of an access token.   Also, I'm only asking whether `ADMISSION` should be a private route, not `ADMISSION_LOGIN` just as I only meant to ask about `PROFILE` before, not `AFE`.  (GitHub seems to show one or more earlier lines for context.)

jschanker (commented):
Including automatically collapsed review comments from before that seem to be inaccessible in certain situations (e.g., if you're not logged in to GitHub).

src/pages/Navgurukul/merakiAdmission/index.js
jschanker: It seems preferable to use `URLSearchParams` here to explicitly handle encoding of reserved characters (e.g., if a last name contains an apostrophe).  So you could do the following:
```suggestion
        const ADMISSIONS_URL = `https://admissions.navgurukul.org`;
        const url = `${ADMISSIONS_URL}/check_duplicate/` +
            new URLSearchParams({
              Name: `${firstName}${middleName}${lastName}`, 
              Number: mobileNumber, 
              Stage: stage
            }).toString();
```
But you'd remove `ADMISSIONS_URL` in place of an import statement if you move it to a config file.

src/pages/Navgurukul/merakiAdmission/index.js
```diff
@@ -0,0 +1,211 @@
...
+      <div className="left-section">
+        <YouTube className={"video-alumni"} videoId={`vuSwndj5cbs`} />;
+        <div className="video-label">Experience of NG Alumni & Graduates</div>
+      </div>
+      <div className="right-section">
+        <div className="test-form">
+          <h2 className="scholarship-test">
+            Software Engineering Scholarship Test
+          </h2>
+          <div className="input-form">
```
jschanker, line 134: For better accessibility (See: https://reactjs.org/docs/accessibility.html#accessible-forms), you can use the [`<label>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label) element for label text for each of the form inputs instead of `<span>`.  If you surround the `<input>` elements with `<label>` and `</label>`, the person will be able to click on the label (First Name, Middle Name, Last Name, or Mobile Number) and the inputs will be put into focus.

src/pages/Navgurukul/index.js
Poonam-Singh-Bagh: Yes.

src/pages/Navgurukul/merakiAdmission/index.js
Poonam-Singh-Bagh: This api will get call with the token which we are getting from android and if we won't get a token from android then only it will get call with token from user/me.

src/pages/Navgurukul/merakiAdmission/index.js
Poonam-Singh-Bagh: Sorry, I didn't get it.

src/pages/Navgurukul/merakiAdmission/index.js
Poonam-Singh-Bagh: Yes.

src/pages/Navgurukul/merakiAdmission/index.js
Poonam-Singh-Bagh: Yeah right, Thanks.

src/routing/index.js
Poonam-Singh-Bagh: Yes, it should be private, I've made it private in the previous commit.

src/pages/Navgurukul/merakiAdmission/index.js
Poonam-Singh-Bagh: `   <label>First Name</label>
                <input
                  type="text"
                  placeholder="Abhi..."
                  value={userDetails.firstName}
                  name="firstName"
                  onChange={changeHandler}
                  required
                />`
               I did like this because wrapping `<input>` with `<label>` caused to bring the label and field on the same line.

src/pages/Navgurukul/merakiAdmission/index.js
jschanker: Sorry, I wasn't being very clear with all my questions, API call?  What I meant is that as per the guidelines in the repository README file (https://github.com/navgurukul/bhanwari-devi#how-to-do-an-api-call-and-manage-state), I thought that all these `axios` API calls would ideally be made from a file `api.js` in a redux folder.  In the case here, the only difference between the call made in this file and that from index.js of Profile was the token that was used.  So we could add a function to api.js like this:

```
export const getProfileData = (token) => {
  return axios({ 
   method: METHODS.GET, 
   url: `${process.env.REACT_APP_MERAKI_URL}/users/me`, 
   headers: { 
     accept: "application/json", 
     Authorization: token, 
   }, 
 })
```

In this call, `getProfileData` is called with a `token` of `userToken` whereas on the Profile page, it's called with the value of `user.data.token`.  More abstraction can be done with redux/saga, but this could be a start.

src/pages/Navgurukul/merakiAdmission/index.js
jschanker: I just meant `firstName`, `middleName`, and `lastName` are all strings, correct?  In that case, you should be able to write line 100 without using `concat` as below:
```
          Name: firstName + " " + middleName + lastName,
```

src/pages/Navgurukul/merakiAdmission/index.js
jschanker: In that case, to keep a connection between the `<label>` and the `<input>`, you can give the `input` an `id` and use the [`for`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#attr-for) attribute.  However, because `for` is a keyword in JavaScript, you do this with `htmlFor` in React.  So for example, you could add an `id` of `firstName` to the input and then the label would be `<label htmlFor = "firstName">`.

src/pages/Navgurukul/merakiAdmission/index.js
Poonam-Singh-Bagh: @jschanker actually we don't know Redux. Neither @komalahire knows nor I. If you can ever get the time to make us understand we'll do it.

jschanker: @Poonam-Singh-Bagh @komalahire Sure, happy to try to help you learn.  Here's a link to some interactive exercises that can help you get started: https://www.freecodecamp.org/learn/front-end-development-libraries/#redux .  And I can try to answer your questions as you go through the lessons and connect them to how you're using Redux here.

Poonam-Singh-Bagh: Thank you @jschanker, where can we communicate? Are you on NavGurukul or Meraki slack?

jschanker:
@Poonam-Singh-Bagh I just logged into Meraki Learn, and I can see a Mentor chat room.  Is that what you mean?

abhishekgupta92:
She is referring to the Slack channel that we have for Meraki. Shall I
share an invite?

jschanker:
Sure, sounds good, thanks!

--
Jason A. Schanker, PhD
Associate Professor
Mathematics and Computer Studies Department
Molloy College
1000 Hempstead Avenue
PO Box 5002
Rockville Centre, NY 11571-5002
Office: Casey 218C
Tel: (516) 323-3430
E-mail: ***@***.***
________________________________

Issue #89, opened 2021-05-24
Change <a> tag to <Link> components


nilshah98:
**Context:**
`a` tags are used to navigate between HTML documents, but they lead to a complete page refresh.
`Link` tag, a component of `react-router-dom` is used to navigate to different page in a react app, more programatically without refreshing the whole page.
Major benefits include, that the state is persisted, since there is no page refresh. Aside from the benefits with such as page rendering and panting.

**Problem:**
In our `Header` component, all the links mentioned in them use `a` tag, so when you navigate using those, it causes the whole page to refresh and start from scratch. Hence, all state is lost, and we need to render state again.

**Fix:**
- [ ] Use `Link` component instead of `a` tag
- [ ] Add `className` of `link` to the `Link` tag to mimic the styles

nilshah98:
Also in `CourseCard` component and `ContinueExercise` component.

PR #230, opened 2022-01-28
State partner dashboard

komalahire:
**Created state partnership dashboard**

@jschanker


src/components/Dashboard/StudentData/index.js
```diff
@@ -60,7 +60,9 @@ function StudentData() {
     }).then((res) => {
       if (
         id == user.data.user.partner_id ||
-        user.data.user.rolesList.indexOf("admin") > -1
+        user.data.user.rolesList.indexOf("admin") > -1 ||
+        user.data.user.rolesList.indexOf("partner_view") > -1 ||
+        user.data.user.rolesList.indexOf("partner_edit") > -1
```
jschanker: Since the logic of checking that one Array contains at least one item from another Array appears multiple times, it may be good to add a function to [utils.js](https://github.com/navgurukul/bhanwari-devi/blob/main/src/common/utils.js) as follows:
```
const hasOneFrom = (arr1, arr2) => arr2.some((item) => arr1.includes(item));
```
Then you could write the following here (after exporting the function from utils.js and importing it from this file):
```suggestion
        hasOneFrom(user.data.user.rolesList, ["admin", "partner_view", "partner_edit"]);
```

src/components/Dashboard/StudentData/index.js
```diff
@@ -280,6 +282,8 @@ function StudentData() {
 
   if (
     user.data.user.rolesList.indexOf("admin") > -1 ||
+    user.data.user.rolesList.indexOf("partner_view") > -1 ||
+    user.data.user.rolesList.indexOf("partner_edit") > -1 ||
```
jschanker: After importing the function:
```suggestion
    hasOneFrom(user.data.user.rolesList, ["admin", "partner_view", "partner_edit"]);
```

src/components/Header/index.js
```diff
@@ -29,9 +28,21 @@ const AuthenticatedHeaderOption = () => {
     });
   }, []);
 
+  const partnerGroupId = user.data.user.partner_group_id;
+
+  const canSpecifyPartnerGroupId =
+    (rolesList.includes("admin") ||
+      rolesList.includes("partner") ||
+      rolesList.includes("partner_view")) &&
```
jschanker: After importing the function:
```suggestion
    hasOneFrom(rolesList, ["admin", "partner", "partner_view"]) &&
```

src/components/Header/index.js
```diff
@@ -29,9 +28,21 @@ const AuthenticatedHeaderOption = () => {
...
+    user.data.user.partner_group_id;
+
   const canSpecifyUserBaseRole = rolesList.indexOf("admin") > -1;
 
-  const canSpecifyPartner = rolesList.includes("partner") && partnerId != null;
+  const canSpecifyPartner =
+    (rolesList.includes("partner") ||
+      // rolesList.includes("partner_view") ||
+      rolesList.includes("partner_edit")) &&
```
jschanker: After importing the function:
```suggestion
    hasOneFrom(rolesList, ["partner", "partner_edit"]) &&
```

src/routing/index.js
```diff
@@ -36,6 +36,11 @@ const Routing = () => {
       <Route exact path={PATHS.PROFILE} component={Profile} />
       <Route exact path={PATHS.OPPORTUNITIES} component={Opportunities} />
       <Route exact path={PATHS.VOLUNTEER} component={VolunteerDashboard} />
+      <Route
+        exact
+        path={PATHS.STATEPARTNER}
+        component={StatePartnerDashboard}
+      />
```
jschanker: Should this be a private route since it requires an access token?

src/pages/StatePartner Dashboard/index.js
```diff
@@ -0,0 +1,143 @@
...
+              <>
+                {i == 0 ? (
+                  <option key={i} value={item.partner_group_name} selected>
+                    {item.partner_group_name}
+                  </option>
+                ) : (
+                  <option key={i} value={item.partner_group_name}>
+                    {item.partner_group_name}
+                  </option>
+                )}
```
jschanker:
```suggestion
              <option
                key={i}
                value={item.partner_group_name}
                selected={i === 0}
              >
                {item.partner_group_name}
              </option>
```

src/components/Header/index.js
```diff
@@ -49,13 +60,21 @@ const AuthenticatedHeaderOption = () => {
...
+            Dashboard
+          </a>
+        </>
+      ) : canSpecifyPartner ? (
         <>
           <a className="item" href={`${PATHS.PARTNERS}/${partnerId}`}>
             Dashboard
           </a>
         </>
       ) : null}
```
jschanker:
```suggestion
      {canSpecifyPartnerGroupId || canSpecifyPartner ? (
        <>
          <a
            className="item"
            href={
              canSpecifyPartnerGroupId
                ? `${PATHS.STATE}/${partnerGroupId}`
                : `${PATHS.PARTNERS}/${partnerId}`
            }
          >
            Dashboard
          </a>
        </>
      ) : null}
```

src/components/Header/index.js
```diff
@@ -49,13 +60,21 @@ const AuthenticatedHeaderOption = () => {
...
-      {canSpecifyPartner ? (
+
+      {canSpecifyPartnerGroupId ? (
+        <>
+          <a className="item" href={`${PATHS.STATE}/${partnerGroupId}`}>
```
jschanker: Maybe use `<Link>` instead of `<a>` as per https://github.com/navgurukul/bhanwari-devi/issues/89 or maybe good to leave as is to leave a good first issue for new contributors

src/routing/index.js
Poonam-Singh-Bagh: Yes

src/pages/StatePartner Dashboard/index.js
Poonam-Singh-Bagh: Getting error when adding `{i === 0 ? "selected" : ""}`

src/pages/StatePartner Dashboard/index.js
```diff
@@ -0,0 +1,157 @@
...
+                <>
+                  <option
+                    key={i}
+                    value={item.partner_group_name}
+                    selected={i === 0 ? "selected" : ""}
+                  >
+                    {item.partner_group_name}
+                  </option>
+                </>
+              );
```
jschanker: The outer `<>` is not necessary here because the JSX you're returning only has one root element `<option>`.  You would need to use a JSX fragment if the `return` statement included adjacent JSX elements such as:
```
return (
  <option>Something</option>
  <option>Something else</option>
);
```
Also, to conditionally select an option using a `selected` attribute, you'd pass it `true` or `false`.  `"selected"` and `""` work because the first value `"selected"` is truthy and the second one of the empty string `""` is falsy but you could just write `i === 0` instead.  (Sorry about my mistake before when I just wrote `{i === 0 ? "selected" : ""}` which won't work.)

One more note is that using `selected` is generally not recommended (see: https://stackoverflow.com/a/21736116), but I see you're using it here because the list of district options will be empty until the district partner data is loaded. 

```suggestion
              return (
                  <option
                    key={i}
                    value={item.partner_group_name}
                    selected={i === 0}
                  >
                    {item.partner_group_name}
                  </option>
              );
```

src/pages/StatePartner Dashboard/index.js
```diff
@@ -0,0 +1,157 @@
...
+    }).then((res) => {
+      setDistrictPartner(res.data);
+      setDistrictPartnerData(res.data.partner_groups_data);
+      setFilterData([...filterData, res.data.partner_groups_data[0]]);
+    });
+  }, []);
+
+  const handleSelcet = (e) => {
+    const stateFilterData = districtPartnerData.filter(
+      (el) => el.partner_group_name == e.target.value
```
jschanker: Using `===` (or `!==`) instead of `==` (and `!=`) is recommended here and anywhere you don't need to do type conversion before the comparison (i.e., if what appears on the left and right of the `===` are expected to be the same type such as here where they should both be strings).
```suggestion
      (el) => el.partner_group_name === e.target.value
```

src/pages/StatePartner Dashboard/index.js
Poonam-Singh-Bagh: We are using selected to display the second option. As `All` is the first option, by default it gets displayed but we want to display the 1st element of the district.

I texted you on slack, asking for help on the Admission feature.

Issue #62, opened 2021-01-26
Remove moment in favour of date-fns


tarun-dugar:
Moment.js is on its way to deprecation. Let's replace it with date-fns.

nilshah98:
Hi @tarun-dugar, @pysaquib can you guys confirm the date format which we are using ?

As far as I can guess by looking at the frontend code, we are using date and time from HTML forms and the follow the below formats -
Date - `YYYY-MM-DD`
Time - `HH:mm`

As for the backend, we are sending it to them as - `YYYY-MM-DDTHH:mm:ss` and adding `Z` at the end
And we are using `ISOString` in backend so removing the `Z` ?

If so, I can start to port to `date-fns`

PR #232, opened 2022-02-02
Feat/python cohort

Poonam-Singh-Bagh:
@jschanker

### DIFF
--- a/src/components/Class/ClassCard/index.js
+++ b/src/components/Class/ClassCard/index.js
@@ -106,6 +106,7 @@ function ClassCard({ item, editClass, enroll, style }) {
 
   //API CALL FOR /users/enrolment to check whether user is enrolled or not
   const checkEnrolled = (pathway_id) => {
+    console.log("pathway_id", pathway_id);
     return axios({
       method: METHODS.POST,
       url: `${process.env.REACT_APP_MERAKI_URL}/users/enrolment?email=${user.data.user.email}&pathway_id=${pathway_id}`,
@@ -115,12 +116,17 @@ function ClassCard({ item, editClass, enroll, style }) {
       },
     })
       .then((res) => {
-        setClassEnrolled(res.data.message);
+        // setClassEnrolled(res.data.message);
+        // setClassEnrolled("not_enrolled");
+        setClassEnrolled("enrolled");
+        // setClassEnrolled("enrolled_but_finished");
       })
       .catch((err) => {
         console.log("err", err);
       });
   };
+
+  console.log("classEnrolled", classEnrolled);
 
   // API CALL FOR enroll class
   const handleSubmit = (Id, indicator) => {
@@ -272,7 +278,7 @@ function ClassCard({ item, editClass, enroll, style }) {
               <label>
                 <input
                   type="checkbox"
-                  align="center"
+                  // align="center"
                   className="cohort-class"
                   onClick={() => {
                     setDeleteCohort(true);
@@ -303,7 +309,7 @@ function ClassCard({ item, editClass, enroll, style }) {
               <label>
                 <input
                   type="checkbox"
-                  align="center"
+                  // align="center"
                   className="cohort-class"
                   onClick={() => {
                     setIndicator(true);
@@ -334,7 +340,7 @@ function ClassCard({ item, editClass, enroll, style }) {
           >
             {classEnrolled === "enrolled" && item.type === "cohort" ? (
               <>
-                <p align="center">
+                <p className="modal-text">
                   You'll be enrolled to a single class of Batch <br /> As you
                   are already enrolled in another Batch.
                 </p>
@@ -411,7 +417,7 @@ function ClassCard({ item, editClass, enroll, style }) {
               <label>
                 <input
                   type="checkbox"
-                  align="center"
+                  // align="center"
                   className="cohort-class"
                   onClick={() => {
                     setIndicator(true);
--- a/src/components/Class/ClassCard/index.js
+++ b/src/components/Class/ClassCard/index.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect, useState } from "react";
 import moment from "moment";
 import { useSelector, useDispatch } from "react-redux";
 import axios from "axios";
@@ -10,18 +10,20 @@ import { toast } from "react-toastify";
 import "react-toastify/dist/ReactToastify.css";
 import Modal from "../../common/Modal";
 import Loader from "../../common/Loader";
+import { check } from "prettier";
 
 toast.configure();
 
 function ClassCard({ item, editClass, enroll, style }) {
   const dispatch = useDispatch();
-  const [enrollShowModal, setEnrollShowModal] = React.useState(false);
-  const [unenrollShowModal, setunenrollShowModal] = React.useState(false);
-  const [showModal, setShowModal] = React.useState(false);
-  const [editShowModal, setEditShowModal] = React.useState(false);
-  const [deleteCohort, setDeleteCohort] = React.useState(false);
-  const [indicator, setIndicator] = React.useState(false);
-  const [loading, setLoading] = React.useState(false);
+  const [enrollShowModal, setEnrollShowModal] = useState(false);
+  const [unenrollShowModal, setunenrollShowModal] = useState(false);
+  const [showModal, setShowModal] = useState(false);
+  const [editShowModal, setEditShowModal] = useState(false);
+  const [deleteCohort, setDeleteCohort] = useState(false);
+  const [indicator, setIndicator] = useState(false);
+  const [loading, setLoading] = useState(false);
+  const [classEnrolled, setClassEnrolled] = useState("");
   const user = useSelector(({ User }) => User);
   const classStartTime = item.start_time && item.start_time.replace("Z", "");
   const classEndTime = item.end_time && item.end_time.replace("Z", "");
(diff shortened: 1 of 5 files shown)


src/components/Class/ClassCard/index.js
```diff
@@ -309,33 +332,73 @@ function ClassCard({ item, editClass, enroll, style }) {
...
-                  className="cohort-class"
-                  onClick={() => {
-                    setIndicator(true);
-                  }}
-                />
-                Enroll all classes of this cohort?
-              </label>
+            {classEnrolled === "enrolled" && item.type === "cohort" ? (
+              <>
+                <p align="center">
```
jschanker: `align` is for styling so it should be done in CSS with [`text-align`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-align).  It's [obsolete in HTML](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/p#attributes).

src/components/Class/ClassCard/index.js
```diff
@@ -309,33 +332,73 @@ function ClassCard({ item, editClass, enroll, style }) {
...
+                      handleSubmit(item.id, false);
+                    }}
+                    className="agree-btn"
+                  >
+                    Enroll
+                  </button>
+                  <button onClick={handleCloseEnroll} className="cancel-btn">
+                    Cancel
+                  </button>
+                </div>
```
jschanker, line 359: Other than a `true/false` change, this is duplicated multiple times.  Could you create a new `Component` from this to avoid repetition?

src/components/Class/index.js
```diff
@@ -235,14 +250,20 @@ function Class({ classToEdit, indicator }) {
     setCheckedState(updatedCheckedState);
   };
 
-  const checkBoxHandler = (e, day, key, field, setField) => {
+  const checkBoxHandler = (e, days, count, key, field, setField) => {
```
jschanker, line 253: Having a parameter variable `days` that shadows the one defined on line 103 with different meanings can be confusing, but if you get rid of the above definition on line 103, which I think you can do if you use my later suggestion, this comment is no longer relevant.

src/components/Class/index.js
```diff
@@ -235,14 +250,20 @@ function Class({ classToEdit, indicator }) {
...
+    const obj = {
+      [count]: days,
+    };
     if (e.target.checked === true) {
```
jschanker:
```suggestion
    if (e.target.checked) {
```

src/components/Class/index.js
```diff
@@ -939,59 +987,62 @@ function Class({ classToEdit, indicator }) {
...
+                        }
+                        value={formFieldsState[OCCURRENCE]}
+                        placeholder="How many recurring classes"
+                        disabled={isEditMode && !indicator ? true : false}
+                        required={
+                          formFieldsState[TYPE] === "cohort" &&
+                          formFieldsState[UNTIL] === ""
+                            ? true
+                            : false
+                        }
```
jschanker: Jab aapke paas ek Boolean condition that matches expression ka value, expression zaroori nahi hai. (Hope my mix of Hindi was correct; main seekhane kee koshish kar raha hoon.)
```suggestion
                        disabled={isEditMode && !indicator}
                        required={
                          formFieldsState[TYPE] === "cohort" &&
                          formFieldsState[UNTIL] === ""
                        }
```

src/components/Class/index.js
```diff
@@ -939,59 +987,62 @@ function Class({ classToEdit, indicator }) {
...
+                        value={formFieldsState[UNTIL]}
+                        placeholder="Until when recurring classes"
+                        disabled={isEditMode && !indicator ? true : false}
+                        required={
+                          formFieldsState[TYPE] === "cohort" &&
+                          formFieldsState[OCCURRENCE] === ""
+                            ? true
+                            : false
+                        }
```
jschanker, line 1015:
```suggestion
                        disabled={isEditMode && !indicator}
                        required={
                          formFieldsState[TYPE] === "cohort" &&
                          formFieldsState[OCCURRENCE] === ""
                        }
```

src/components/Class/index.js
```diff
@@ -765,6 +806,7 @@ function Class({ classToEdit, indicator }) {
                           checkBoxHandler(
                             e,
                             "MO",
+                            1,
```
jschanker, line 809: If you include the numbers representing the days, are the abbreviations also necessary?

src/components/Class/index.js
```diff
@@ -291,7 +312,22 @@ function Class({ classToEdit, indicator }) {
...
+          while (i <= 7) {
+            incrementedDate = moment(incrementedDate).add(1, "days")._d;
+            let Day = incrementedDate.toString().split(" ")[0];
+            if (days[weekDday[0]] === Day) {
+              newDate = incrementedDate;
+              break;
+            }
+            i = i + 1;
+          }
+          formFields[fieldName] = moment.utc(newDate).format("YYYY-MM-DD");
```
jschanker, line 329: Since [moment.js](https://momentjs.com/docs/#/-project-status/) is now a legacy project and you're planning to migrate away from it (https://github.com/navgurukul/bhanwari-devi/issues/62), I'd recommend avoiding additional usage of it as possible.

In this case, if you want to get the next weekday from `weekDday[0]` and format in UTC, you could do something like the following (after importing `addDays` from `date-fns` and `format` and `utcToZonedTime` from `date-fns-tz`).  For more information about the conversion to UTC below, see [this StackOverflow answer](https://stackoverflow.com/a/63227335).  Please review as I wasn't sure if this was what you were trying to do: 
```suggestion
        const weekDday = Object.values(day);
        if (fieldName === "start_time") {
          const weekDdayIndex = ["SU", "MO", "TU", "WE", "TH", "FR", "SA"].indexOf(weekDday[0]);
          const dayIndex = new Date(value).getDay();
          const daysUntilNext = weekDdayIndex - dayIndex + (dayIndex < weekDdayIndex ? 0 : 7);
          formFields[fieldName] = format(
              utcToZonedTime(addDays(new Date(value), daysUntilNext), 'UTC'),
              'yyyy-MM-dd',
              { timeZone: 'UTC' }
          );
```

src/components/Class/index.js
```diff
@@ -97,6 +100,16 @@ function Class({ classToEdit, indicator }) {
...
 
+  const days = {
+    MO: "Mon",
+    TU: "Tue",
+    WE: "Wed",
+    TH: "Thu",
+    FR: "Fri",
+    SA: "Sat",
+    SU: "Sun",
+  };
```
jschanker, line 111: I think you could remove this if you use the previous suggestion I make as I only saw it referenced on line 323.

src/components/Class/index.js
```diff
@@ -10,6 +10,7 @@ import Loader from "../common/Loader";
 import Form from "../common/form";
 import { METHODS } from "../../services/api";
 import "./styles.scss";
+import { setDay } from "date-fns";
```
jschanker, line 13: This import is unused; the `setDay` used as the function for the `useState` hook is shadowing this `setDay`.  But if you're changing this line and implementing my later suggestion, you can instead import [`addDays`](https://date-fns.org/v2.28.0/docs/addDays).  Also, you would need to add imports of [`format`](https://www.npmjs.com/package/date-fns-tz#format) and [`utcToZonedTime`](https://www.npmjs.com/package/date-fns-tz#user-content-utctozonedtime) from `date-fns-tz`.

src/components/Class/index.js
Poonam-Singh-Bagh: Yes. That abbreviations are being send to backend

src/components/Class/index.js
Poonam-Singh-Bagh: I'm using it to change the date according to the days

src/components/Class/index.js
jschanker: Lekin, kyaa back end sankhyaon se determine dinon ke naam nahi kar sakta?  I think it might be better if there were a constants file that contained a mapping from the day numbers to their abbreviations.  This constants file could then be shared by the front and back ends so if the abbreviations were to change (e.g., MO to MON), you'd only need to change them in one place.

src/components/Class/index.js
jschanker: (Fixed a copy and paste error in the suggestion.)

src/components/Class/ClassCard/index.js
Poonam-Singh-Bagh: New Component for what?

src/components/Class/index.js
Poonam-Singh-Bagh: @jschanker It's not able to resolve `date-fns-tz` even after [installing](https://www.npmjs.com/package/date-fns-tz)

src/components/Class/index.js
Poonam-Singh-Bagh: I don't know about backend

src/components/Class/ClassCard/index.js
jschanker: Something like this:
```
function ConfirmEnroll(props) {
  return (
    <div className="wrap">
      <button
        onClick={() => {
          handleSubmit(props.item.id, props.applyAll);
        }}
        className="agree-btn"
      >
        Enroll
      </button>
      <button onClick={handleCloseEnroll} className="cancel-btn">
        Cancel
      </button>
    </div>
  );
}
```

And then either
```
<ConfirmEnroll item={item} applyAll={true} />
```
or
```
<ConfirmEnroll item={item} applyAll={false} />
```
depending on whether you're passing `true` or `false` as the second argument to `handleSubmit`.

src/components/Class/index.js
jschanker: Not sure about this as I was able to get the imports to work in a CodeSandbox, which handles dependency installation for you, but this may be related (https://github.com/marnusw/date-fns-tz/issues/133).

PR #233, opened 2022-02-03
Admission route as private route

Poonam-Singh-Bagh:
@jschanker

### DIFF
--- a/src/routing/index.js
+++ b/src/routing/index.js
@@ -43,7 +43,7 @@ const Routing = () => {
         component={NavgurukulIntroduce}
       />
       <Route exact path={PATHS.REDIRECT} component={RedirectComponent} />
-      <Route exact path={PATHS.ADMISSION} component={Admission} />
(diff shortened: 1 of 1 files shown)


jschanker:
@Poonam-Singh-Bagh I think the issue is this line here because you're comparing `userToken` against the string `"undefined"` instead of `undefined` itself.  So, since `userToken !== "undefined"` should always be `true` (`undefined !== "undefined"`), it'll always use `userToken` and never `user.data.token`.
https://github.com/navgurukul/bhanwari-devi/blob/fbc3dafd7536d4c19a75339915a40bcd7afdeb18/src/pages/Navgurukul/merakiAdmission/index.js#L30

Poonam-Singh-Bagh:
@jschanker, Yes, I'll remove the string. But that is not the problem.
If the user is logged in on Android, web shouldn't ask for login on web but it's asking

PR #234, opened 2022-02-08
3 items of header shifted to left side

Poonam-Singh-Bagh:
1. Created 2 functions named as `AuthenticatedLeftHeaderOption` and `PublicLeftMenuOption` and added items that should be on the left side
2. Changed the class name, of Course, Class, and Mentor from item to left-item in `AuthenticatedHeaderOption`  function and make it display none in desktop size and display block in phone size
3. Make `AuthenticatedLeftHeaderOption` function's items display none in phone size

@jschanker

### DIFF
--- a/src/components/Header/index.js
+++ b/src/components/Header/index.js
@@ -60,13 +60,13 @@ const AuthenticatedHeaderOption = () => {
       <a className="item" href={PATHS.ADMISSION}>
         Admission
       </a>
-      <a className="item" href={PATHS.COURSE}>
+      <a className="left-item" href={PATHS.COURSE}>
         Courses
       </a>
-      <a className="item" href={PATHS.MENTOR}>
+      <a className="left-item" href={PATHS.MENTOR}>
         Mentor
       </a>
-      <a className="item" href={PATHS.CLASS}>
+      <a className="left-item" href={PATHS.CLASS}>
         Classes
       </a>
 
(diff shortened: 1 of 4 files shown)


src/pages/Navgurukul/merakiAdmission/index.js
```diff
@@ -27,7 +27,7 @@ function Admission() {
       url: `${process.env.REACT_APP_MERAKI_URL}/users/me`,
       headers: {
         accept: "application/json",
-        Authorization: userToken !== "undefined" ? userToken : user.data.token,
+        Authorization: userToken !== undefined ? userToken : user.data.token,
```
jschanker: When the `Token` key is not present in the `localStorage`, `localStorage.getItem("Token");` will `return` `null` but `null !== undefined` is `true`.  So this is a case where you'll want to use `!=` instead because `null != undefined` will be `false`.  Alternatively, you can compare with `null` instead of `undefined`.  (See: https://developer.mozilla.org/en-US/docs/Web/API/Storage/getItem)
```suggestion
        Authorization: userToken != undefined ? userToken : user.data.token,
```

src/components/Header/index.js
```diff
@@ -123,6 +136,14 @@ const PublicMenuOption = () => {
   );
 };
 
+const PublicLeftMenuOption = () => {
+  return (
+    <a className="item" href={PATHS.COURSE}>
+      Courses
+    </a>
+  );
+};
```
jschanker, line 132: I think it might be good to create an object that maps the path keys to human-readable names.  E.g.,
```
const NAMES = {
  COURSE: "Courses",
  MENTOR: "Mentors",
  CLASS: "Classes"
};
```
You can add the names for the other links as well.  The point is that there could be an object constant like this for each supported language (e.g., Hindi, Tamil, etc.) so if the user selects a different language, it would be a simple matter of reassigning `NAMES`.  

Then I'd also recommend creating a component that contains a single navigation menu option like this:
```
const MenuOption = (props) => {
  return (
    <a className="item" href={PATHS[props.id]}>
      {NAMES[props.id]}
    </a>
  );
};
```
So for example, `AuthenticatedLeftHeaderOption` would look like this 
```
const AuthenticatedLeftHeaderOption = () => {
  return (
    <>
      {["COURSE", "MENTOR", "CLASS"].map((id) => (
        <MenuOption id={id} />
      ))}
    </>
  );
};
```
This would make it easy to say change an `a` to `Link`.

Poonam-Singh-Bagh: But in `AuthenticatedHeaderOption` we are using another className. There is 2 classNames `item` and `left-item`, how about that? if you're asking to do the same with this `AuthenticatedHeaderOption` function as well.

PR #235, opened 2022-02-11
fix: Login via Redirect token

jschanker:
Allows login of already registered user via token entered in query string of `/redirect`, used for logging into Admissions (https://github.com/navgurukul/bhanwari-devi/pull/233)

1. Added `sendToken` method to `User` redux api.js, which uses the user token to get the user's profile data via a GET request to [/users/me](https://github.com/navgurukul/sansaar/blob/main/lib/routes/users.js#L445).
2. Modified User redux saga.js to call either this `sendGoogleUserData` or `sendUserData` depending on whether data has `idToken` required for request to [/users/auth/google](https://github.com/navgurukul/sansaar/blob/main/lib/routes/users.js#L66).
3. Modified logic of OnlyLoggedIn.js to get `Token` from `localstorage` and if present and the User is not already authenticated, to dispatch an [`ON_USER_SIGN_INTENT`](https://github.com/navgurukul/bhanwari-devi/blob/59b1aebe6e33902857ef491062207f543467ba4c/src/components/User/redux/action.js#L10) action with this `token` data.  `localstorage` would have its `Token` key set to the value of the `token` parameter in the query string of a request made to `/redirect`.
4. Made Admissions route private for testing fix.

**(Currently a draft but creating pull request for testing purposes.)**

@Poonam-Singh-Bagh

### DIFF
--- a/src/components/User/redux/api.js
+++ b/src/components/User/redux/api.js
@@ -21,6 +21,21 @@ export const sendGoogleUserData = (userData, tokens) => {
   });
 };
 
+
+/**
+ * Sends id-token to Meraki back-end to get profile data of registered user.
+ */
+export const sendToken = (userData, tokens) => {
+  return axios({
+      method: METHODS.GET,
+      url: `${process.env.REACT_APP_MERAKI_URL}/users/me`,
+      headers: {
+        accept: "application/json",
+        Authorization: userData.token,
+      },
+    });
+}
+
 export const updateUser = (userData, tokens) => {
   return axios({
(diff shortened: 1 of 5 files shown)


Poonam-Singh-Bagh:
@jschanker Did you test this in your local? I tested this with aws but it's still redirecting to login page.

jschanker:
> @jschanker Did you test this in your local? I tested this with aws but it's still redirecting to login page.

@Poonam-Singh-Bagh It doesn't work yet.  I've been testing it on AWS because I haven't been able to test locally.

PR #242, opened 2022-02-26
Style customize

Poonam-Singh-Bagh:


src/theme/theme.js
```diff
@@ -2,140 +2,129 @@ import { createTheme } from "@mui/material/styles";
 
 const theme = createTheme({
   palette: {
-    type: "dark",
+    type: "light",
```
jschanker, line 5: @Poonam-Singh-Bagh From https://mui.com/guides/migration-v4/, <q>The `theme.palette.type` key was renamed to `theme.palette.mode`, to better follow the "dark mode" term that is usually used for describing this feature.</q> 
```suggestion
    mode: "light",
```

src/theme/theme.js
```diff
@@ -2,140 +2,129 @@ import { createTheme } from "@mui/material/styles";
...
+    default: {
+      // main: "#ffffff",
+      // contrastText: "#000000",
+      light: "#0066ff",
+      main: "#0044ff",
+      // dark: will be calculated from palette.secondary.main,
+      contrastText: "#ffcc00",
+    },
     primary: {
+      //Green
```
jschanker, line 15: Instead of specifying `main`, `light`, and `dark` directly for `green` and the other colors, you may want to only specify the value of `main` with a shade of a default color:

E.g., you could write

`import { green } from '@mui/material/colors';`

```
primary: {
  main: green[500]
}
```
As noted at the bottom of the section https://mui.com/customization/color/#playground, <q>If you are using the default primary and / or secondary shades then by providing the color object, `createTheme()` will use the appropriate shades from the material color for main, light and dark.</q>

You can also experiment with color schemes for accessibility purposes here: https://mui.com/customization/color/#official-color-tool.

src/theme/theme.js
Poonam-Singh-Bagh: Sure. Thanks.

PR #247, opened 2022-03-15
New designed footer

Poonam-Singh-Bagh:
@jschanker


src/components/Footer/styles.js
```diff
@@ -0,0 +1,19 @@
+import { makeStyles } from "@mui/styles";
+
+const useStyles = makeStyles((theme) => ({
```
jschanker, line 3: I noticed that according to https://mui.com/styles/basics/, `makeStyles` is now a legacy solution for styling so it may be best to avoid it for a new project:  
> It depends on [JSS](https://cssinjs.org/) as a styling solution, which is not used in the @mui/material anymore, deprecated in v5

src/components/Footer/index.js
```diff
@@ -1,15 +1,153 @@
...
+            <Grid
+              xs={12}
+              md={6}
+              sx={{
+                pr: { sm: "none", md: "17px" },
+              }}
+            >
+              <Typography
+                color="text.primary"
+                sx={{ textAlign: { sm: "none", md: "right" } }}
```
jschanker: `none` is not a valid value for `textAlign` in CSS.  (I'm not positive about MUI.)  Did you want to use `left` instead?
```suggestion
                sx={{ textAlign: { sm: "left", md: "right" } }}
```

src/components/Footer/index.js
```diff
@@ -1,15 +1,153 @@
...
+            <Grid xs={12} md={6} sx={{ pl: { sm: "none", md: "10px" } }}>
+              <Typography color="text.primary">
+                Legal & Privacy Policy
+              </Typography>
+            </Grid>
+            <Grid
+              xs={12}
+              md={6}
+              sx={{
+                pr: { sm: "none", md: "17px" },
```
jschanker: Similarly `0` for right padding instead of `none`?
```suggestion
                pr: { sm: 0, md: "17px" },
```

src/components/Footer/index.js
```diff
@@ -1,15 +1,153 @@
...
+                  loading="lazy"
+                />
+              </Box>
+              <Box className={classes.image}>
+                <img
+                  src={require("./asset/twitter.svg")}
+                  alt="social media"
+                  loading="lazy"
+                />
+              </Box>
```
jschanker: You may want to consider making a Component out of footer icons that would look something like this:
```
function FooterIcon(props) {
  return (
    <Box className={classes.image}>
      <img
           src={require("./asset/" + props.name + ".svg")}
           alt={props.alt || "image of " + props.name}
           loading="lazy"
      />
    </Box>
  );
}
```
And then you could rewrite this as something like:
```suggestion
              {['logo', 'meraki', 'facebook', 'linkedIn', 'twitter'].map((imgName) => 
                  <FooterIcon name={imgName} />)}
```

src/components/Footer/index.js
```diff
@@ -1,15 +1,153 @@
...
+                  Now on Playstore
+                </Typography>
+              </Box>
+            </Box>
+          </Grid>
+        </Grid>
+        <Divider variant="string" sx={{ pt: "25px" }} />
+        <Box>
+          <Grid container spacing={2} sx={{ m: "30px 0px 30px 0px" }}>
+            <Grid xs={12} md={6} sx={{ pl: { sm: "none", md: "10px" } }}>
```
jschanker:
```suggestion
            <Grid xs={12} md={6} sx={{ pl: { sm: 0, md: "10px" } }}>
```

src/components/Footer/index.js
```diff
@@ -1,15 +1,153 @@
...
   return (
-    <div className="footer">
-      <Link className="footer-link" to={PATHS.PRIVACY_POLICY}>
-        Privacy Policy
-      </Link>
-    </div>
+    <Box maxWidth="false" bgcolor="primary.light">
+      <Container maxWidth="xl">
+        <Grid container spacing={2} sx={{ mt: "40px" }}>
+          <Grid xs={12} md={4} sx={{ pl: { sm: "none", md: "16px" } }}>
```
jschanker:
```suggestion
          <Grid xs={12} md={4} sx={{ pl: { sm: 0, md: "16px" } }}>
```

src/components/Footer/index.js
Poonam-Singh-Bagh: Yes.

src/components/Footer/index.js
Poonam-Singh-Bagh: I cannot add logo and meraki in that because I gave a different class for them. So I did it as you told me with social media only.

src/components/Footer/styles.js
Poonam-Singh-Bagh: So what can we use for styling?

PR #249, opened 2022-03-15
Profile

Poonam-Singh-Bagh:
@jschanker


src/pages/Profile/index.js
```diff
@@ -53,39 +64,49 @@ function Profile() {
...
   return (
-    <div className="profile-container">
-      <div className="profile">
-        <div>
-          <img className="profile-img" src={userData.profile_picture} />
-        </div>
-        <div className="profile-details">
+    <Container maxWidth="lg" sx={{ pt: 20 }}>
+      <Grid container spacing={2}>
+        <Grid item xs={12} md={6} align={isActive ? "center" : "right"}>
```
jschanker, line 69: Between breakpoints of 600 and 900 pixels, the image appears on the right and the text appears on the left but on different lines.  This is probably undesirable.  Also, you probably don't want to use `align` as this will produce this obsolete attribute in the resulting HTML.  Finally, as you seem to be using this type of aligning for a number of different elements, maybe it'd be useful to create a file with these styles.  For example:
```
const smAlignStyle =
  (smallAlign, notSmallAlign) =>
  ({ theme }) => ({
    [theme.breakpoints.down('sm')]: {
      textAlign: smallAlign,
    },
    [theme.breakpoints.up('sm')]: {
      textAlign: notSmallAlign,
    },
  });

export const centerSmRightNonSmAlignStyle = smAlignStyle('center', 'right');
export const centerSmLeftNonSmAlignStyle = smAlignStyle('center', 'left');
```

Then you could import these constants and have the following stylized components for use:
```
const LeftCenterAlignedGrid = styled(Grid)(centerSmLeftNonSmAlignStyle);
const RightCenterAlignedGrid = styled(Grid)(centerSmRightNonSmAlignStyle);
```
```suggestion
        <RightCenterAlignedGrid item xs={12} sm={6}>
```

src/pages/Profile/index.js
```diff
@@ -53,39 +64,49 @@ function Profile() {
...
+          <Avatar
+            alt="Remy Sharp"
+            src={userData.profile_picture}
+            sx={{ width: 100, height: 100 }}
+          />
+        </Grid>
```
jschanker, line 75:
```suggestion
        </RightCenterAlignedGrid>
```

src/pages/Profile/index.js
```diff
@@ -53,39 +64,49 @@ function Profile() {
...
+        <Grid item xs={12} md={6} align={isActive ? "center" : "none"}>
```
jschanker, line 76: Similar comments to above.  Remember also that `none` is not a valid value for `align`:
```suggestion
        <LeftCenterAlignedGrid item xs={12} sm={6}>
```

src/pages/Profile/index.js
```diff
@@ -53,39 +64,49 @@ function Profile() {
...
-              Edit Name
-            </button>
+            <Button onClick={() => setIsEditing(true)}>
+              {!isActive && "Edit Profile"}
+            </Button>
           )}
-        </div>
-      </div>
-    </div>
+        </Grid>
```
jschanker, line 107:
```suggestion
        </LeftCenterAlignedGrid>
```

jschanker (commented):
There are probably some better (and shorter) names for my suggested styles/stylized Component constants.  For references about styled and the potential replacements for the legacy `makeStyles`, see: https://mui.com/system/styled/, https://mui.com/customization/how-to-customize/, https://mui.com/guides/interoperability/, and https://mui.com/system/basics/.

src/pages/Profile/index.js
Poonam-Singh-Bagh: Why do we need to do this? We can give class as well right?

jschanker: That's the solution I wanted to use here as well.  Unfortunately, `makeStyles` is legacy, and while not an issue in this case, the values of properties assigned by ordinary CSS rules with class selectors could take lower precedence over conflicting default styles of MUI components without workarounds such as including `!important`, making the selectors more specific by say adding classes, or using `injectFirst`.  (See: https://mui.com/guides/interoperability/#css-injection-order.)  It could be fine in this case.

PR #255, opened 2022-03-19
New login page

komalahire:


src/pages/Login/index.js
```diff
@@ -42,6 +44,9 @@ function Login(props) {
     // dispatch(userActions.onUserUpdate(referrer));
   }
 
+  const classes = useStyles();
+  const isActive = useMediaQuery("(max-width:600px)");
```
jschanker, line 48: Instead of hard-coding `600px` in each of these files, I think it'd be better to be referring to the theme constants (e.g., `theme.breakpoints.values.sm`).  (See: https://mui.com/customization/breakpoints/#custom-breakpoints.)  This way if you decide to change the `theme` breakpoint, it'd be updated across files.  Unfortunately, I don't know how you can access this constant directly from this Component without passing it down or creating some sort of [Context](https://reactjs.org/docs/context.html).  So, this is what I'd suggest as a workaround.  This workaround is probably good to guard against the very unlikely possibility that the defaults unexpectedly change in a later major version update of MUI anyway:
1. In the constant.js file, add:
```
export const breakpoints = {
  values: {
    xs: 0,
    sm: 600,
    md: 900,
    lg: 1200,
    xl: 1536,
  },
};
```
2. In the `theme.js` file, `import` this constant and then include it as a property with this value:
```
import { breakpoints } from '../constant'

createTheme(theme, {
  breakpoints,
  palette: {
  ...
}
```
3. Then in this file (and any file where you use `isActive`), you could `import` `breakpoints` from the `constant.js` file and refer to it:

```suggestion
  const isActive = useMediaQuery("(max-width:" + breakpoints.values.sm + "px)");
```

What do you think?

src/pages/Login/styles.js
```diff
@@ -0,0 +1,24 @@
+import { makeStyles } from "@mui/styles";
+
+const useStyles = makeStyles((theme) => ({
```
jschanker, line 3: I noticed that according to https://mui.com/styles/basics/, `makeStyles` is now a legacy solution for styling so it may be best to avoid it for a new project:  
> It depends on [JSS](https://cssinjs.org/) as a styling solution, which is not used in the @mui/material anymore, deprecated in v5

src/pages/Login/index.js
```diff
@@ -75,21 +80,51 @@ function Login(props) {
...
-    </div>
+    <>
+      <Container className={classes.merakiLogin} maxWidth="lg">
+        <Grid container spacing={2}>
+          <Grid item xs={12} ms={6} md={6}>
+            <Container maxWidth="sm">
+              <Typography
+                sx={{ pt: { xs: "none", md: 24 } }}
+                variant="h4"
+                align={isActive ? "center" : "left"}
```
jschanker, line 91: You probably don't want to use `align` as this will produce this obsolete attribute in the resulting HTML.  Also, as you seem to be using this type of aligning for a number of different elements, maybe it'd be useful to create a file with these styles.  For example:
```
const smAlignStyle =
  (smallAlign, notSmallAlign) =>
  ({ theme }) => ({
    [theme.breakpoints.down('sm')]: {
      textAlign: smallAlign,
    },
    [theme.breakpoints.up('sm')]: {
      textAlign: notSmallAlign,
    },
  });

export const centerSmRightNonSmAlignStyle = smAlignStyle('center', 'right');
export const centerSmLeftNonSmAlignStyle = smAlignStyle('center', 'left');
```

Then you could import these constants and have the following stylized component for use:
```
const LeftCenterAlignedTypography = styled(Typography)(centerSmLeftNonSmAlignStyle);
```

I also mention a similar suggestion in the Profile PR: https://github.com/navgurukul/bhanwari-devi/pull/249#discussion_r830564262.

src/pages/Login/styles.js
komalahire: Then if we want to give CSS  then how can we give?

src/pages/Login/index.js
komalahire: Yes.  i have did this change and it's very useful.  Will add for home page as well.

src/pages/Login/styles.js
jschanker: > Then if we want to give CSS then how can we give?

If you're only using the given style once, you could use the `sx` prop (https://mui.com/customization/how-to-customize/#use-the-sx-prop).  If it's to be applied everywhere, you'd probably make it part of the theme.  Otherwise, if it's meant to apply to some but not all elements (e.g., when you'd use `class`es), you'll probably want to go with `styled`.  I give an example of this use in the below comment (See: https://mui.com/customization/how-to-customize/#2-reusable-style-overrides).

You could also create a [Higher-order component](https://reactjs.org/docs/higher-order-components.html) to simulate multiple class membership:

```
function withClasses(Component, classes) {
  return classes.reduce((styledComponent, className) => styled(styledComponent)(className), Component);
}

const LeftCenterAlignedTypography = withClasses(Typography, [centerSmLeftNonSmAlignStyle]);
```

Poonam-Singh-Bagh: The style we are giving is for only that particular component, I don't think the same style will work for other components too. So in that case what should we do? Inline CSS using `sx` props?
Like I need to use `paddingTop: 5 and width: 100` 3 times in my component and I'm also supposing I don't have to use them on other components so now I'm using styles.js file, but you're suggesting to use `sx` props in this case, if I understood right?

jschanker: Yes, the underlying principle here is to avoid repetition of values that you always expect to stay the same.  So for example, if `paddingTop` were to change from `5` to `10`, you should only be changing it in one place.

So, in this case, if you have three instances of the same component with the same styles, you could assign the component to a variable as follows (which you'd export/import if you were using across multiple files):
```
const ____ = <Component sx = {{paddingTop: 5, width: 100}}>...</Component>;
```
If on the other hand, these Components had additional differing styles, you could assign an object with the common styles to a variable and then use it as part of the `sx` value as follows:
```
const commonStyles = {paddingTop: 5, width: 100};
<Component sx = {{...commonStyles, color: red}}>...</Component>;
<Component sx = {{...commonStyles, margin: 10}}>...</Component>;
```

PR #305, opened 2022-04-13
Team page

bunnie26:
Team page completed


src/pages/TeamPage/index.js
```diff
@@ -0,0 +1,348 @@
...
+  let newObj = {};
+  var keys = Object.keys(obj);
+  keys.sort(function (a, b) {
+    return Math.random() - 0.5;
+  });
+  keys.forEach(function (k) {
+    newObj[k] = obj[k];
+  });
+  return newObj;
+}
```
jschanker: There are a few issues with this function.  The first is that the properties of objects with non-numerical keys don't have an inherent order.  Instead, you're attempting to shuffle the **insertion order** of the key-value pairs and are relying on `Object.keys` to retain this order.  But if the object were an Array (keys are the indices), nothing would get changed as `Object.keys` would list the indices in increasing order.  Similarly, if you have some keys that are nonnegative integers and others not, the nonegative integer keys will always appear first.  For example:
```
const obj = {};
obj['foo'] = 'something';
obj[42] = 'bar';
console.log(Object.keys(obj)); // logs 42. then foo
```
Even if you only have keys that are strings, the situation with `Object.keys` preserving insertion order is more complicated prior to ES2020 (See: https://stackoverflow.com/a/30919039).

Another issue is that the shuffling algorithm will be biased and have time complexity O(nlgn) (assuming the sort has this complexity but this is an implementation detail: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) instead of O(n), which will make it a little more inefficient for larger Arrays.

Finally, you're only using the Array returned by `Object.keys` so it would make more sense to return that rather than the object itself.

So to address these issues, I'd use the standard [Fisher-Yates shuffle algorithm](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle) which takes an Array as an input, shuffles it, and returns the shuffled Array as an output all in linear time.  Additionally, I would put this function in `utils.js` as it is a general function that's not specific to this component and could be used elsewhere so this could be the function to export:
```
const shuffleArr = (arr) => {
  for(let i = arr.length-1; i >= 0; i--) {
    const newIndex = Math.floor(Math.random() * (i + 1));
    [arr[i], arr[newIndex]] = [arr[newIndex], arr[i]]
  }
  return arr;
};
```

src/pages/TeamPage/index.js
```diff
@@ -0,0 +1,348 @@
...
+              className={
+                !isActive
+                  ? `${classes.team_infoCardContaier} team-info-cards-container`
+                  : `${classes.team_infoResponsiveContainer} team-info-cards-container`
+              }
+              sx={{ marginTop: 4 }}
+            >
+              <Grid container>
+                {Object.keys(shuffleObject(team)).length ? (
+                  Object.keys(shuffleObject(team)).map((item) => {
```
jschanker: No need to shuffle twice and using the above suggestion, you could do the following:
```suggestion
                {Object.keys(team).length ? (
                  shuffleArray(Object.keys(team)).map((item) => {
```

jschanker, line 254: No need to shuffle twice and using the above suggestion, you could do the following:
```suggestion
                {Object.keys(team).length ? (
                  shuffleArray(Object.keys(team)).map((item) => {
```

jschanker (commented):
Some initial comments

src/pages/TeamPage/index.js
```diff
@@ -0,0 +1,348 @@
...
+  const [members, setMembers] = useState({
+    teamMembers: true,
+    volunteers: false,
+  });
+  const classes = useStyles();
+  const isActive = useMediaQuery("(max-width:" + breakpoints.values.sm + "px)");
+  const isActiveIpad = useMediaQuery("(max-width:1300px)");
+  // console.log(breakpoints.values.sm);
+  console.log(isActive);
+  function Popup(props) {
```
jschanker, line 52: When you mouse over someone's picture on a browser whose width is greater than or equal to 600px, you don't see any text because it's white and the background is white.  The names below people's pictures are also white and are not aligned over their titles.

src/components/common/SocialMediaIcons/Facebook.js
```diff
@@ -0,0 +1,46 @@
+import React from "react";
+
+function Facebook() {
```
jschanker, line 3: Each of these icons has a commonality that should be factored out into a `SocialMediaIcon` component with appropriate properties.  This will also ensure consistency between them if something were to change (e.g., `width/height`)

src/pages/TeamPage/index.js
```diff
@@ -0,0 +1,349 @@
...
+import "./style.css";
+import { breakpoints } from "../../theme/constant";
+import LinkedIn from "../../components/common/SocialMediaIcons/LinkedIn";
+import Twitter from "../../components/common/SocialMediaIcons/Twitter";
+
+/**
+ *
+ * @param {Array} arr
+ * mutates Array inplace
+ */
```
jschanker, line 25: For JSDoc comments, put the description at the top and include an `@return` annotation when something is returned.  Also, remember to put this function in utils.js and import it here as it's a general function that might be used elsewhere.
```suggestion
/**
 * Mutates Array in place
 * @param {Array} arr the Array to shuffle
 * @return {Array} the shuffled Array
 */
```

src/pages/TeamPage/index.js
```diff
@@ -0,0 +1,349 @@
...
+}
+
+function TeamPage() {
+  useEffect(() => {
+    axios({
+      url: `https://anandpatel504.github.io/tarabai-shinde/data/ng_team.json`,
+    }).then((res) => {
+      setTeam(res.data);
+    });
+  }, []);
```
jschanker, line 41: Redux again, time permitting.  Also, you might want to use a constant for the URL as it seems like something that could change.

src/pages/TeamPage/index.js
jschanker: I'm not sure if this was addressed yet, but I still can't see any of the team members' names or descriptions without highlighting the text.

src/pages/TeamPage/index.js
```diff
@@ -0,0 +1,349 @@
...
+            <a
+              href={props.twitter}
+              target="_blank"
+              className={classes.team_socialIcon}
+            >
+              <Twitter />
+            </a>
+          ) : (
+            <></>
+          )}
```
jschanker, line 81: Use `ExternalLink` here?
https://github.com/navgurukul/bhanwari-devi/blob/47aa4663606dd7ecfd2ab7ab1b93b720f1d029f0/src/components/common/ExternalLink/index.js#L3-L5
Using this and some factoring of commonality:
```suggestion
          {[
                  { Comp: LinkedIn, href: props.linkedin },
                  { Comp: Twitter, href: props.twitter },
                ].map(
                  (socialMedia) =>
                    socialMedia.href && (
                      <ExternalLink
                        href={socialMedia.href}
                        className={classes.team_socialIcon}
                      >
                        <socialMedia.Comp />
                      </ExternalLink>
                    )
                )}
```

src/pages/TeamPage/index.js
```diff
@@ -0,0 +1,349 @@
...
+  }
+  return (
+    <>
+      <CssBaseline />
+      <div>
+        <Container
+          maxWidth="lg"
+          style={
+            isActive ? { padding: 0, marginTop: "24px" } : { marginTop: "40px" }
+          }
```
jschanker, line 106: You can also use [responsive values](https://mui.com/material-ui/react-grid/#responsive-values) instead of `style` for the margins.  But for padding, this might not work as intended as setting it to `0` for `xs` and not `sm` means it'll always be `0`.
```suggestion
          mt={{ xs: "24px", sm: "40px" }}
          style={ isActive ? { padding: 0 } : { } }
```

src/pages/TeamPage/index.js
```diff
@@ -0,0 +1,349 @@
...
+                  </Box>
+                  <Typography className={"d"} variant="h4" align="right">
+                    Supporters
+                  </Typography>
+                </div>
+                <hr color="primary" className={classes.team_hrline} />
+              </Box>
+              <Typography
+                variant="body1"
+                align={!isActive ? "center" : "left"}
```
jschanker, line 161:
```suggestion
                textAlign={{xs: "left", sm: "center"}}
```

jschanker (commented):
Some more comments

src/pages/TeamPage/index.js
```diff
@@ -0,0 +1,349 @@
...
+                      (condition === "volunteers" &&
+                        team[item].Association === "Volunteer") ||
+                      (condition === "teamMembers" &&
+                        team[item].Association !== "Volunteer")
+                    ) {
+                      if (
+                        team[item].Photo &&
+                        team[item].Name &&
+                        team[item].Content.length &&
+                        team[item].Content &&
```
jschanker, line 269: If `team[item].Content` is `undefined` or `null`, there would be an error if you wrote it the other way around. 
```suggestion
                        team[item].Content &&
                        team[item].Content.length &&
```

src/pages/TeamPage/index.js
```diff
@@ -0,0 +1,349 @@
...
+                                  : "right"
+                              }
+                              content={
+                                <Popup
+                                  Name={
+                                    team[item].Name || "Awaiting Member's Name"
+                                  }
+                                  Content={
+                                    (team[item].Content.length &&
+                                      team[item].Content) ||
```
jschanker, line 290: Same issue as before for `undefined` or `null` `Content`.
```suggestion
                                    (team[item].Content &&
                                      team[item].Content.length) ||
```

src/pages/TeamPage/index.js
```diff
@@ -0,0 +1,349 @@
...
+    });
+  }, []);
+  const [team, setTeam] = useState([]);
+  const [members, setMembers] = useState({
+    teamMembers: true,
+    volunteers: false,
+  });
+  const classes = useStyles();
+  const isActive = useMediaQuery("(max-width:" + breakpoints.values.sm + "px)");
+  const isActiveIpad = useMediaQuery("(max-width:1300px)");
```
jschanker, line 49: Is this used?  If it is, you should still be getting the `1300px` from your theme as you do for the `600px` for `isActive`.
```suggestion
```

src/pages/TeamPage/index.js
```diff
@@ -0,0 +1,349 @@
...
+                        borderBottom: "3px solid #48a145",
+                      }
+                    : {}
+                }
+              >
+                Core Team
+              </Typography>
+              <Typography
+                onClick={() => {
+                  setMembers({ teamMembers: false, volunteers: true });
```
jschanker, line 225:
```suggestion
                  setShowTeam(false);
```

src/pages/TeamPage/index.js
```diff
@@ -0,0 +1,349 @@
...
+              !isActive
+                ? classes.team_containerTopSpace
+                : `${classes.team_responsiveContainer}`
+            }
+          >
+            <Grid container style={{ justifyContent: "center" }}>
+              <Typography
+                onClick={() => {
+                  setMembers({ volunteers: false, teamMembers: true });
+                }}
```
jschanker, line 205:
```suggestion
                  setShowTeam(true);
```

src/pages/TeamPage/index.js
```diff
@@ -0,0 +1,349 @@
...
+      url: `https://anandpatel504.github.io/tarabai-shinde/data/ng_team.json`,
+    }).then((res) => {
+      setTeam(res.data);
+    });
+  }, []);
+  const [team, setTeam] = useState([]);
+  const [members, setMembers] = useState({
+    teamMembers: true,
+    volunteers: false,
+  });
```
jschanker, line 46: Since `teamMembers` and `volunteers` are always opposite Boolean values, do you really need both of them?
```suggestion
  const [showTeam, setShowTeam] = useState(true);
```

src/pages/TeamPage/index.js
```diff
@@ -0,0 +1,349 @@
...
+                  setMembers({ teamMembers: false, volunteers: true });
+                }}
+                variant="subtitle1"
+                className={
+                  !isActive
+                    ? classes.team_selector
+                    : classes.team_MobileSelector
+                }
+                style={
+                  members.volunteers
```
jschanker, line 234:
```suggestion
                  !showTeam
```

src/pages/TeamPage/index.js
```diff
@@ -0,0 +1,349 @@
...
+                  setMembers({ volunteers: false, teamMembers: true });
+                }}
+                variant="subtitle1"
+                className={
+                  !isActive
+                    ? classes.team_selector
+                    : classes.team_MobileSelector
+                }
+                style={
+                  members.teamMembers
```
jschanker, line 213:
```suggestion
                  showTeam
```

src/pages/TeamPage/index.js
```diff
@@ -0,0 +1,349 @@
...
+                  shuffleArray(Object.keys(team)).map((item) => {
+                    const condition = members.teamMembers
+                      ? "teamMembers"
+                      : "volunteers";
+                    if (
+                      (condition === "volunteers" &&
+                        team[item].Association === "Volunteer") ||
+                      (condition === "teamMembers" &&
+                        team[item].Association !== "Volunteer")
+                    ) {
```
jschanker, line 264: I think it might be a little clearer to write this code excerpt as something like this although the specific name of the non-volunteer association is not really used here (i.e., it could be anything that's not `Volunteer`):
```suggestion
                    const association = showTeam ? "Team" : "Volunteer";
                    if (team[item].Association === association) {
```

However, the association name `Volunteer` should probably be a constant as it's dependent on having the same name (and currently case: if value changed to `volunteer`, the above code wouldn't work) as the `Association` from the JSON file.

jschanker (commented):
More comments

src/pages/TeamPage/index.js
bunnie26: Actually popup comes but it overflows to right as we around 600px when we switch it to mobile and reload the browser it will come below the name .
[
image
](url)

jschanker: @bunnie26 This is what I'm seeing:
image

When selecting the text, I see this:
image

When inspecting the element, the problematic generated CSS rule is:
```
.jss9 {
    color: white;
    text-decoration: none;
}
```
And if I remove the `color: white` declaration, I see this as I should (the alignment seems to be good on this run):
image

PR #313, opened 2022-04-15
state dashboard

Poonam-Singh-Bagh:


src/pages/StatePartner Dashboard/Dashboard/index.js
```diff
@@ -0,0 +1,253 @@
...
+    setValue(e.target.value);
+  };
+
+  const selectRegion = (e) => {
+    let regionPartner;
+    region.filter((item) => {
+      if (item.partner_group_name === e.target.value) {
+        regionPartner = item.partner_groups_data.partner_groups_data;
+      }
+    });
```
jschanker: I wouldn't use `filter` here as the purpose of that method is to create a new subarray from the existing one.  Additionally, assuming there is only one item with the `partner_group_name` target value, you can stop as soon as you find it rather than continuing through the Array as `filter` would do.  In this case, you could use the following: 
```suggestion
    const regionItem = region.find((item) => item.partner_group_name === e.target.value); 
    const regionPartner = regionItem && regionItem.partner_groups_data.partner_groups_data;
```

src/pages/StatePartner Dashboard/Dashboard/index.js
```diff
@@ -0,0 +1,253 @@
...
+    setFilterData([regionPartner[0]]);
+  };
+
+  const selecetDistrict = (e) => {
+    let destrictPartner;
+    regionDistrictPartner.filter((item) => {
+      if (item.partner_group_name === e.target.value) {
+        destrictPartner = [item];
+      }
+    });
```
jschanker: Assuming there is exactly one such item, the following would be equivalent:
```suggestion
    const districtPartner = [
      regionDistrictPartner.find(item => item.partner_group_name === e.target.value)
    ];
```

src/pages/StatePartner Dashboard/Dashboard/index.js
```diff
@@ -0,0 +1,253 @@
...
+      if (
+        res.data.partner_groups_data[0].partner_group_name.includes("_REGION")
+      ) {
+        setFilterData([
+          res.data.partner_groups_data[0].partner_groups_data
+            .partner_groups_data[0],
+        ]);
+      } else {
+        setFilterData([res.data.partner_groups_data[0]]);
+      }
```
jschanker:
```suggestion
      const partnerGroups = res.data.partner_groups_data;
      const includesRegion = partnerGroups[0].partner_group_name.includes("_REGION"); 
      setDistrictPartner(res.data);
      setDistrictPartnerData(partnerGroups);
      setRegionData(partnerGroups);
      
      if (includesRegion) {
        setRegionDistrictPartner(partnerGroups[0].partner_groups_data.partner_groups_data);
        setFilterData([partnerGroups[0].partner_groups_data.partner_groups_data[0]]);
      } else {
        setFilterData(partnerGroups[0]);
      }
```

src/pages/StatePartner Dashboard/Dashboard/index.js
```diff
@@ -0,0 +1,253 @@
...
+    });
+  }, [stateId]);
```
jschanker: Time permitting, this would be good to do with Redux again.

jschanker (commented):
Few initial comments

src/pages/StatePartner Dashboard/Dashboard/index.js
Poonam-Singh-Bagh: This is urgent now so I did like this... Later when we'll convert it into MUI we'll call this API in redux.

PR #330, opened 2022-04-26
added language to the card

Poonam-Singh-Bagh:


komalahire (dismissed):

src/components/Class/ClassCard/index.js
```diff
@@ -195,7 +195,7 @@ function ClassCard({ item, editClass, enroll, style }) {
             className={classes.spacing}
           >
             {languageMap[item.type]}
-            {item.enrolled && (
+            {item.enrolled == true && (
```
jschanker: Was it not working properly without `== true` here?
Cases:
`item.enrolled = true` : `item.enrolled && <i>Enrolled</i>` should evaluate to `<i>Enrolled</i>`
`item.enrolled = false`, `undefined`, or `null`:  `item.enrolled && <i>Enrolled</i>` should evaluate to `false`, `undefined` or `null`, respectively, each of which should cause nothing to be displayed.

On the other hand, if `item.enrolled` can be a **string** `"true"` or `"false"`, neither of these will work because as strings, `"true"` and `"false"` are both truthy meaning that Enrolled would be displayed in each case without the `== true` whereas with `== true`, Enrolled won't show in either case because in JavaScript,  `"true" == true` is `false`!  So, if you have to handle both cases, you could do the following:
```suggestion
            {String(item.enrolled) === "true" && (
```
If you know it'll never be a Boolean `true`, you could leave off the `String`.

Poonam-Singh-Bagh: Reverting it. For some reason, we added it but we don't need now so doing it again`item.enrolled && `

komalahire (approved):

PR #290, opened 2022-04-08
Review of New Design

jschanker:

PR #334, opened 2022-04-27
design feedback

Poonam-Singh-Bagh:


src/pages/TeamPage/index.js
```diff
@@ -58,10 +58,10 @@ function TeamPage() {
       item.Association !== null &&
       item.Photo !== null &&
       item.Name !== null &&
-      item.Content !== null &&
+      item.Content.length > 1 &&
```
jschanker: Should this be `> 0` or maybe `item.Content.filter(s => s.trim()).length > 0` (if you remove the "Awaiting" string message as suggested below) to ask that the Array of non-whitespace items be nonempty?  I'm not sure why this data is being stored in an Array in the file as opposed to an ordinary string in the first place or why there are new line Array items?

src/pages/TeamPage/index.js
```diff
@@ -285,67 +284,65 @@ function TeamPage() {
...
+                              window.screen.availWidth < 650
+                                ? "bottom"
+                                : "right"
+                            }
+                            content={
+                              <Popup
+                                Name={item.Name || "Awaiting Member's Name"}
+                                Content={
+                                  (item.Content.length && item.Content) ||
+                                  "Awaiting content from team member"
```
jschanker: Few questions/suggestions:

1. Why is `item.Content` used in `item.Content.length && item.Content`?  If `item.Content` has to be an `Array`, then this will always be a truthy value so it'll be unnecessary.  If `item.Content` can be `null` or `undefined`, this will cause an error because `item.Content.length` is being evaluated first.
2. I would recommend against using a hard-coded string `"Awaiting content from team member"` here that's repeated in the Popup component for determining the color of the text because it's easy to accidentally change it in one place and forget to change it in the other whereby the loading grey color won't be applied.  It'd probably be better to check in the Popup component if the `length` of the `props.Content` is `0` and then display this message, accordingly.
3. Finally, I don't think it's good for the Content to be able to be an Array or a String depending on whether the content is loading or not as again, it'll be possible that you might do something specific to Arrays in the Popup that would then cause an error if used on strings.  For example, if you wanted to `filter` it to see if the length of the Array of non-whitespace items is greater than 0 or use `join('<br \>')` to separate the Content items with line breaks `<br \>`.
4. Instead of using English text here, might it be better to use constants dependent on language that could dynamically update when the user's language is changed?  For example, if you go to https://www.google.com/?hl=hi, you'll get a user interface in Hindi whereas if you go to https://www.google.com/?hl=en, it'll update to English.  This would be my suggestion in other places as well (e.g., header item text).

src/pages/TeamPage/index.js
```diff
@@ -241,7 +240,7 @@ function TeamPage() {
                         fontWeight: "bold",
                         borderBottom: "3px solid #48a145",
                       }
-                    : {}
+                    : { color: "#9c9999" }
```
jschanker, line 243: Use as a theme color for consistency across pages?  (This is also used here for example: https://github.com/navgurukul/bhanwari-devi/blob/432c85b158044d7dd48b260935994b5dda9188fd/src/pages/StatePartner%20Dashboard/style.scss#L128-L130)

src/pages/TeamPage/index.js
```diff
@@ -285,67 +284,65 @@ function TeamPage() {
...
-                                  className={`${classes.team_cardDetails} card-details`}
+                                <Typography
+                                  variant="body1"
+                                  className={classes.team_cardTitle}
+                                  style={
+                                    !isActive ? {} : { textAlign: "center" }
+                                  }
+                                >
+                                  {item.Name}
+                                </Typography>
```
jschanker, line 334: Recommend earlier [ResponsiveTypography suggestion](https://github.com/navgurukul/bhanwari-devi/pull/290#discussion_r853781939):
```
function ResponsiveTypography(props) {
  return <Typography textAlign={{xs: "center", sm: "left"}} {...props}/>
}
```
```suggestion
                                <ResponsiveTypography
                                  variant="body1"
                                  className={classes.team_cardTitle}
                                >
                                  {item.Name}
                                </ResponsiveTypography>
```

src/pages/TeamPage/index.js
Poonam-Singh-Bagh: Ohh sorry! My bad.

src/pages/TeamPage/index.js
Poonam-Singh-Bagh: 1. If `item.Content.length` then it will display the content `item.Content`.  That's why it has written `item.Content.length && item.Content`
2. If the content has length `item.Content.length` then it will display the content `item.Content` or else it will display the string.
I didn't get 3 & 4. What are you talking about?

src/pages/TeamPage/index.js
Poonam-Singh-Bagh: This will get changed soon so that's why I didn't refer to it.

src/pages/TeamPage/index.js
Poonam-Singh-Bagh: What about sx that you've mentioned here [ResponsiveTypography suggestion](https://github.com/navgurukul/bhanwari-devi/pull/290#discussion_r853781939)

src/pages/TeamPage/index.js
jschanker: 1.  Sorry, disregard this point, I was focusing on the possibility of the errors that would occur if `item.Content` is `null` or `undefined`, which we're assuming won't happen.
2. My point here is that you're also referring to the hard-coded string in the below for styling so if you decide to change the message that's shown on line 303 (to something like "Description N/A", you'd also have to remember to change it below or this text wouldn't appear as grey): https://github.com/navgurukul/bhanwari-devi/blob/5c16f440990690ea5eca00d386fcfd2f1d693563/src/pages/TeamPage/index.js#L107-L111
My recommendation here would be to pass `item.Content` (regardless of whether it's an empty Array or not), and then assign a constant like `const isContentAvailable = props.Content.filter(item => item.trim()).length === 0`.  Then you could replace the above lines with `style={isContentAvailable ? {color: grey} : {}}`  
3. With the current setup, if `item.Content.length` is `0`, then `props.Content` will be a **string** "Awaiting content from team member".  If its length is greater than 0, then `props.Content` will be an Array.  So `props.Content` can be a `string` or an `Array`.  What this means is that if the description were say an Array of lines and you wanted it to appear with line breaks in the HTML, you might do something like this: `props.Content.join('<br />')`.  But if `props.Content` can also be the string "Awaiting content from team member", this would cause the code to crash as `join` cannot be called on a string.  You could always add an `if` statement to check that it's an Array, but that will unnecessarily complicate the code.
4. If you did something like this:
`const Msg = {};`
And in en.js, you had an object assignment like this:
```
Msg.en = {
  "TEAM_HEADING": "Meraki Team",
  "AWAITING_CONTENT": "Awaiting content from team member",
};
```
And in hi.js, you had an object assignment like this:
```
Msg.hi = {
  "TEAM_HEADING": "मेराकी टीम",
  "AWAITING_CONTENT": "(Hindi translation for Awaiting content from team member)",
};
```

Then you could write `Msg[currentLanguage].TEAM_HEADING` or `Msg[currentLanguage].AWAITING_CONTENT` and when you switch assignments from `currentLanguage = 'en'` to say `currentLanguage = 'hi'`, the text would all switch from English to Hindi.

src/pages/TeamPage/index.js
jschanker: The component there had an `sx` property so I kept it, but here it does not so I didn't include it.  The `ResponsiveTypography` component allows you to remove the common alignment style from each `Typography` one that you're converting so you'd remove something like this:
```
                                                  align={isActive ? "center" : "left"}
```
And then keep all of the other specified properties.

src/pages/TeamPage/index.js
jschanker: Actually, since you're filtering out empty content Arrays here:
https://github.com/navgurukul/bhanwari-devi/blob/cd36d5f8e04585d2606e7b57fc66ef328722fdd7/src/pages/TeamPage/index.js#L62
will it even be possible to see `"Awaiting content from team member"` now?  Sorry if I'm asking questions about stuff that will change later.

Poonam-Singh-Bagh: No worries. That's absolutely fine. No it won't show this content now.

komalahire (approved):

PR #347, opened 2022-05-03
updated lang

kartiks26:
A clear and concise description. Use `#` to refer to github issues and PR

1. Upgraded packages
3. ...


src/components/PathwayExercise/index.js
```diff
@@ -246,19 +259,25 @@ function PathwayExercise() {
...
+                >
+                  {availableLang.map((lang) => {
+                    return <MenuItem value={lang}>{Lang[lang]}</MenuItem>;
+                  })}
+                </Select>
+              ) : (
+                <MenuItem value={availableLang[0]}>
+                  {Lang[availableLang[0]]}
+                </MenuItem>
+              )}
```
jschanker: Was this meant to be reversed (`<Select>` for when there are multiple available languages) to match the above?  If so, you should create a new Component to avoid code duplication, which as illustrated here can easily lead to inadvertent inconsistencies.

src/components/PathwayExercise/index.js
```diff
@@ -212,19 +219,25 @@ function PathwayExercise() {
...
+                  onChange={(e) => {
+                    setLanguage(e.target.value);
+                  }}
+                  variant="standard"
+                >
+                  {availableLang.map((lang) => {
+                    return <MenuItem value={lang}>{Lang[lang]}</MenuItem>;
+                  })}
+                </Select>
+              )}
```
jschanker: I recommend using a constant for the language menu so you can avoid duplication (Whether it's 1 language or multiple ones, the following should work):
```
const langMenu = availableLang.map((lang) => <MenuItem value={lang}>{Lang[lang]}</MenuItem>);
```
Also, to avoid additional code duplication, I recommend you use a component here, so it'd look something like this:
```
function languageSelectMenu() {
  const langMenu = availableLang.map((lang) =><MenuItem value={lang}>{Lang[lang]}</MenuItem>);
  return (
    availableLang.length === 1 ? 
      langMenu :
      <Select
        disableUnderline
        value={language}
        IconComponent={() => null}
        onChange={(e) => {
          setLanguage(e.target.value);
        }}
        variant="standard"
      >
        langMenu
      </Select>
  )
}
```

src/components/PathwayExercise/index.js
```diff
@@ -114,12 +116,17 @@ function PathwayExercise() {
...
       .catch((err) => {
         console.log("error");
       });
   }, []);
-
+  const Lang = {
+    en: "English",
+    mr: "Marathi",
+    hi: "Hindi",
+  };
```
jschanker: Can you import the language map constant from here: https://github.com/navgurukul/bhanwari-devi/blob/08eaeab3030736341d3b7a3685c9477e3bb40de6/src/pages/CourseContent/languageMap.js#L1-L7 instead even if it currently contains a couple of languages (Tamil and Telugu) for which there are no translations?

komalahire (approved):

PR #369, opened 2022-05-07
feat: Notify user on rolesList change (with polling)

jschanker:
Users who are logged in can now go to `/user/update` and click on a button to start polling for changes (every 5 minutes) to their roles.  As long as they stay logged in (and leave the app open), they will be notified within the polling interval of 5 minutes when their `rolesList` has been changed.  If they exit the app and return later, the polling will resume.  If they allow notifications, they will receive a system notification when the change has occurred.  Regardless, they will receive a toast within the app when this occurs and their local storage will be updated.

1. Added unlisted private route `/user/update` where the user can initiate the polling request.  Clicking the button dispatches an action that initiates the polling request here: https://github.com/navgurukul/bhanwari-devi/blob/85751a7b28d25651f66bc7f2cb232c2ecaf87d41/src/components/User/redux/saga.js#L30-L53
The above generator repeatedly makes a request to get the profile data and compares it with the current user data.  If it changes, it dispatches a `ON_SERVER_USER_UPDATE` action with the updated data and stops polling.
2. Updated `storage.js` file so that when the polling action is initiated and when new data is received (`ON_SERVER_USER_UPDATE`) or when a polling request has been initiated (`POLL_FOR_SERVER_USER_UPDATE`), `localStorage` will be updated, accordingly (https://github.com/navgurukul/bhanwari-devi/blob/85751a7b28d25651f66bc7f2cb232c2ecaf87d41/src/services/storage.js#L18-L30
By adding a `waitingForUpdate` property whose value is `true` when polling is initiated (and `false` when a change has been observed), the user is able to exit the app and come back to it and have the polling resume.
3. Updated the App component to dispatch the `POLL_FOR_SERVER_USER_UPDATE` action that causes polling to start if the `waitingForUpdate` property is `true`.  This is what resumes the polling if the app is reloaded since it loads the `User` state from `localStorage`.  Also, if it changes from `true` to `false`, then the data has been updated so the user will be notified by a toast and also by the built-in Notification API (if the user has allowed it).
4. Factored out selectors in `selectors.js` file to avoid code duplication and to allow for changes to the structure of the `User` object:
https://github.com/navgurukul/bhanwari-devi/blob/85751a7b28d25651f66bc7f2cb232c2ecaf87d41/src/components/User/redux/selectors.js#L1-L4
These selectors can be imported to simplify the inputs to `useSelector` calls (referencing `User`).

@Poonam-Singh-Bagh


jschanker:
Users who are logged in can now go to `/user/update` and click on a button to start polling for changes (every 5 minutes) to their roles.  As long as they stay logged in (and leave the app open), they will be notified within the polling interval of 5 minutes when their `rolesList` has been changed.  If they exit the app and return later, the polling will resume.  If they allow notifications, they will receive a system notification when the change has occurred.  Regardless, they will receive a toast within the app when this occurs and their local storage will be updated.

1. Added unlisted private route `/user/update` where the user can initiate the polling request.  Clicking the button dispatches an action that initiates the polling request here: https://github.com/navgurukul/bhanwari-devi/blob/85751a7b28d25651f66bc7f2cb232c2ecaf87d41/src/components/User/redux/saga.js#L30-L53
The above generator repeatedly makes a request to get the profile data and compares it with the current user data.  If it changes, it dispatches a `ON_SERVER_USER_UPDATE` action with the updated data and stops polling.
2. Updated `storage.js` file so that when the polling action is initiated and when new data is received (`ON_SERVER_USER_UPDATE`) or when a polling request has been initiated (`POLL_FOR_SERVER_USER_UPDATE`), `localStorage` will be updated, accordingly (https://github.com/navgurukul/bhanwari-devi/blob/85751a7b28d25651f66bc7f2cb232c2ecaf87d41/src/services/storage.js#L18-L30
By adding a `waitingForUpdate` property whose value is `true` when polling is initiated (and `false` when a change has been observed), the user is able to exit the app and come back to it and have the polling resume.
3. Updated the App component to dispatch the `POLL_FOR_SERVER_USER_UPDATE` action that causes polling to start if the `waitingForUpdate` property is `true`.  This is what resumes the polling if the app is reloaded since it loads the `User` state from `localStorage`.  Also, if it changes from `true` to `false`, then the data has been updated so the user will be notified by a toast and also by the built-in Notification API (if the user has allowed it).
4. Factored out selectors in `selectors.js` file to avoid code duplication and to allow for changes to the structure of the `User` object:
https://github.com/navgurukul/bhanwari-devi/blob/85751a7b28d25651f66bc7f2cb232c2ecaf87d41/src/components/User/redux/selectors.js#L1-L4
These selectors can be imported to simplify the inputs to `useSelector` calls (referencing `User`).

@Poonam-Singh-Bagh

PR #374, opened 2022-05-10
Feat/searchbar

Poonam-Singh-Bagh:


komalahire (approved):

src/components/SearchBar/SearchCourse/index.js
```diff
@@ -0,0 +1,220 @@
...
+  useEffect(() => {
+    dispatch(courseActions.getCourses());
+  }, [dispatch]);
+
+  useEffect(() => {
+    dispatch(pathwayActions.getPathways());
+  }, [dispatch]);
+
+  const handleSearchChange = (e) => {
+    history.push(`?search=${e.target.value}`);
```
jschanker, line 43: If a person types in p, y, t, h, o, then n, then they will have ?search=p, ?search=py, ?search=pyt, ?search=pyth, ?search=pytho, ?search=python in their history.  Is this desirable?  Also, if they hit the back button, the matches won't update.  So for example, if the user goes back to the page with the search for p, they'll still see the results for python instead of any course with a p.

PR #386, opened 2022-05-14
Assessment

Poonam-Singh-Bagh:
Assessment feature on Exercise

1. Created 2 components `Assessment` and `AssessmentContent` for Displaying the Assessment 
2. Called `Assessment` component in ExerciseContent

@jschanker


src/components/PathwayExercise/ExerciseContent/Assessment/index.js
```diff
@@ -0,0 +1,211 @@
+import React, { useState } from "react";
+import { Container, Box, Typography, Paper, Button } from "@mui/material";
+import useStyles from "./styles";
+import get from "lodash/get";
+import DOMPurify from "dompurify";
+
+const headingVarients = {};
+
+[Typography, "h2", "h3", "h4", "h5", "h6"].forEach(
```
jschanker, line 23: Were these all be supposed to be changed to `Typography` components, or were we still keeping all but the `<h1>` as ordinary HTML headings?

src/components/PathwayExercise/ExerciseContent/Assessment/index.js
```diff
@@ -0,0 +1,211 @@
...
+                          correct={correct}
+                          submit={submit}
+                          setSubmitDisable={setSubmitDisable}
+                        />
+                      ))
+                  : null}
+              </>
+            );
+          }
+        })}
```
jschanker:
```suggestion
      {data && submit &&
        data.map((content) => {
          const dataArr = content.value?.[correct ? "correct" : "incorrect"]; 
          return content.component === "output" && dataArr.map((content, index) => (
            <AssessmentContent
              content={content}
              index={index}
              correct={correct}
              submit={submit}
              setSubmitDisable={setSubmitDisable}
            />            
          ));
        })}
```

src/components/PathwayExercise/ExerciseContent/Assessment/index.js
```diff
@@ -0,0 +1,211 @@
...
+  const [submit, setSubmit] = useState();
+  const [submitDisable, setSubmitDisable] = useState();
+
+  const submitAssessment = () => {
+    setSubmit(true);
+    if (answer == solution) {
+      setCorrect(true);
+    } else {
+      setCorrect(false);
+    }
```
jschanker:
```suggestion
    setCorrect(answer == solution);
```

src/components/PathwayExercise/index.js
```diff
@@ -84,8 +147,18 @@ function NavigationComponent({
...
+              (exercise.content_type === "class_topic" &&
+                `${require("./asset/classtype.svg")}`) ||
+              (exercise.content_type === "exercise" &&
+                `${require("./asset/contentTypeSelectd.svg")}`)
+            : (exercise.content_type === "assessment" &&
+                `${require("./asset/assessment.svg")}`) ||
+              (exercise.content_type === "class_topic" &&
+                `${require("./asset/classtype.svg")}`) ||
+              (exercise.content_type === "exercise" &&
+                `${require("./asset/contenttype.svg")}`)
```
jschanker: Lines 137-165 seem to be duplicated on 367-386 with the exception of pushing to the history.  Additionally, there is  some duplication when mapping `content_type`s to image paths.  As such, I'd make the following two recommendations.  First, create a new Component definition.  Then in the Component definition, use an object to map the names to the paths.  So, it'd look something like this:
```
function ExerciseImage(props) {
  const contentTypeMap = {
    assessment: props.selected ? 'assessmentSelected' : 'assessment',
    class_topic: 'classtype',
    exercise: props.selected ? 'contentTypeSelectd' : 'contenttype',
  };
  return (
    <img
      onClick={() => {
        props.onClick();
        setExerciseId(index);
      }}
      src={require('./asset/' + contentTypeMap[props.contentType] + '.svg')}
      loading="lazy"
      className={classes.contentImg}
    />
  );
}
```
Then you could replace lines 137-165 with the following:
```
      <ExerciseImage
        onClick={() => {
          history.push(
            interpolatePath(PATHS.PATHWAY_COURSE_CONTENT, {
              courseId: params.courseId,
              exerciseId: index,
              pathwayId: params.pathwayId,
            })
          );
        }}
        selected={exerciseId == index}
        contentType={exercise.content_type}
      />
```
while lines 367-386 could be replaced by:
```
                          <ExerciseImage selected={exerciseId == index} contentType={exercise.content_type} />
```

src/components/PathwayExercise/index.js
```diff
@@ -295,8 +368,18 @@ function PathwayExercise() {
...
+                                    `${require("./asset/contentTypeSelectd.svg")}`)
+                                : (exercise.content_type === "assessment" &&
+                                    `${require("./asset/assessment.svg")}`) ||
+                                  (exercise.content_type === "class_topic" &&
+                                    `${require("./asset/classtype.svg")}`) ||
+                                  (exercise.content_type === "exercise" &&
+                                    `${require("./asset/contenttype.svg")}`)
                             }
                             loading="lazy"
                             className={classes.contentImg}
```
jschanker: As noted in the earlier comment, remove the image in favor of:
```
                          <ExerciseImage selected={exerciseId == index} contentType={exercise.content_type} />
```

src/components/PathwayExercise/ExerciseContent/Assessment/index.js
```diff
@@ -0,0 +1,211 @@
...
+      return (
+        <Box sx={{ mt: "32px" }}>
+          <Typography variant="h6" align="center">
+            Output
+          </Typography>
+
+          <Box
+            sx={{
+              mt: "32px",
+              bgcolor: correct ? "#E9F5E9" : "#FFE5E3",
```
jschanker: I'd recommend getting these colors from the theme, so you don't have to replace all occurrences if the theme colors were to ever change.  If you always want them to match `success` and `error` colors, you could use the following:
```suggestion
              bgcolor: correct ? "success.light" : "error.light",
```

src/components/PathwayExercise/ExerciseContent/Assessment/styles.js
```diff
@@ -0,0 +1,25 @@
+import { makeStyles } from "@mui/styles";
+
+const useStyles = makeStyles((theme) => ({
+  option: {
+    boxShadow:
+      "0px 1px 2px #48A145, 0px 2px 1px #48A145, 0px 1px 5px #48A145 !important",
```
jschanker:
```suggestion
      `0px 1px 2px ${theme.palette.primary.main}, 0px 2px 1px ${theme.palette.primary.main}, 0px 1px 5px ${theme.palette.primary.main} !important`,
```

src/components/PathwayExercise/ExerciseContent/Assessment/styles.js
```diff
@@ -0,0 +1,25 @@
...
+  },
+  // correctAnswer: {
+  //   boxShadow:
+  //     "0px 1px 2px #48A145, 0px 2px 1px #48A145, 0px 1px 5px #48A145 !important",
+  //   background: "#E9F5E9",
+  // },
+  correctAnswer: {
+    boxShadow:
+      "0px 1px 2px #48A145, 0px 2px 1px #48A145, 0px 1px 5px #48A145 !important",
```
jschanker:
```suggestion
      `0px 1px 2px ${theme.palette.success.main}, 0px 2px 1px ${theme.palette.success.main}, 0px 1px 5px ${theme.palette.success.main} !important`,
```

src/components/PathwayExercise/ExerciseContent/Assessment/styles.js
```diff
@@ -0,0 +1,25 @@
...
+    background: "#E9F5E9 !important",
+  },
+  inCorrectAnswer: {
+    boxShadow:
+      "0px 1px 2px #F44336, 0px 2px 1px #F44336, 0px 1px 5px #F44336 !important",
```
jschanker:
```suggestion
      `0px 1px 2px ${theme.palette.error.main}, 0px 2px 1px ${theme.palette.error.main}, 0px 1px 5px ${theme.palette.error.main} !important`,
```

src/components/PathwayExercise/ExerciseContent/Assessment/styles.js
```diff
@@ -0,0 +1,25 @@
...
+  },
+  // correctAnswer: {
+  //   boxShadow:
+  //     "0px 1px 2px #48A145, 0px 2px 1px #48A145, 0px 1px 5px #48A145 !important",
+  //   background: "#E9F5E9",
+  // },
+  correctAnswer: {
+    boxShadow:
+      "0px 1px 2px #48A145, 0px 2px 1px #48A145, 0px 1px 5px #48A145 !important",
+    background: "#E9F5E9 !important",
```
jschanker:
```suggestion
    background: ${theme.palette.success.light} !important",
```

src/components/PathwayExercise/ExerciseContent/Assessment/styles.js
```diff
@@ -0,0 +1,25 @@
...
+  },
+  inCorrectAnswer: {
+    boxShadow:
+      "0px 1px 2px #F44336, 0px 2px 1px #F44336, 0px 1px 5px #F44336 !important",
+    background: "#FFE5E3 !important",
```
jschanker:
```suggestion
    background: `${theme.palette.error.light} !important`,
```

src/components/PathwayExercise/ExerciseContent/Assessment/index.js
```diff
@@ -0,0 +1,211 @@
...
+const headingVarients = {};
+
+[Typography, "h2", "h3", "h4", "h5", "h6"].forEach(
+  (Name, index) =>
+    (headingVarients[index + 1] = (data) => (
+      <Name
+        className="heading"
+        dangerouslySetInnerHTML={{ __html: data }}
+        {...(index === 0 ? { component: "h1", variant: "h6" } : {})}
+      />
```
jschanker: There are a number of these components where you're dangerously setting the inner HTML.  As noted by its name, this property needs to be used very carefully when the user can have control over it in some way.  Since you want to try to avoid the possibility of forgetting to call `DOMPurify.sanitize` (as you may have here?) and since you may want to include other safeguards in the future for preventing against XSS attacks, it'd be a good idea to create some sort of component like the following:
```
function UnsafeHTML(props) {
  const {html, Container, ...otherProps} = props;
  const sanitizedHTML = DOMPurify.sanitize(html);
  return <Container {...otherProps} dangerouslySetInnerHTML={{ __html: sanitizedHTML}} /> 
}
```

So assuming you're keeping this as is (`<h_>` for everything except `<h1>`), you could then use this as follows:
```suggestion
      <UnsafeHTML
        Container={Name}
        className="heading"
        html={data}
        {...(index === 0 ? { component: "h1", variant: "h6" } : {})}
      />
```

src/components/PathwayExercise/ExerciseContent/Assessment/index.js
```diff
@@ -0,0 +1,211 @@
...
+              mt: "32px",
+              bgcolor: correct ? "#E9F5E9" : "#FFE5E3",
+              p: "16px",
+              borderRadius: "8px",
+            }}
+          >
+            <Typography
+              variant="body1"
+              dangerouslySetInnerHTML={{ __html: text }}
+            />
```
jschanker:
```suggestion
            <UnsafeHTML
              Container={Typography}
              variant="body1"
              html={text}
            />
```

src/components/PathwayExercise/ExerciseContent/Assessment/index.js
```diff
@@ -0,0 +1,211 @@
...
+        <Box
+          sx={{
+            p: "16px 0",
+            borderRadius: "8px",
+          }}
+        >
+          <Typography
+            variant="body1"
+            dangerouslySetInnerHTML={{ __html: text }}
+          />
```
jschanker:
```suggestion
          <UnsafeHTML
            Container={Typography}
            variant="body1"
            html={text}
          />
```

src/components/PathwayExercise/ExerciseContent/Assessment/index.js
```diff
@@ -0,0 +1,211 @@
...
+            bgcolor: "#F7F7F7",
+            p: "16px 16px 22px 16px",
+            borderRadius: "8px",
+          }}
+        >
+          <Typography
+            sx={{ m: "2rem 0" }}
+            variant="body1"
+            dangerouslySetInnerHTML={{ __html: text }}
+          />
```
jschanker:
```suggestion
          <UnsafeHTML
            Container={Typography}
            sx={{ m: "2rem 0" }}
            variant="body1"
            html={text}
          />
```

jschanker (commented):
May include additional follow-up comments later when I get a chance.

src/components/PathwayExercise/ExerciseContent/Assessment/index.js
Poonam-Singh-Bagh: I added one more line after if else so can't do it now

src/components/PathwayExercise/ExerciseContent/Assessment/styles.js
Poonam-Singh-Bagh: Screenshot from 2022-05-16 20-17-42

It's showing this error

src/components/PathwayExercise/ExerciseContent/Assessment/index.js
Poonam-Singh-Bagh: It's not working, It's throwing arrow on `const dataArr = content.value?[correct ? "correct" : "incorrect"]; `
So I did like this `const dataArr = content.value && correct ? content.value.correct : content.value.incorrect`

BTW how does this work?

jschanker: Sorry, I had left out the `.` before.  It should be:
```
          const dataArr = content.value?.[correct ? "correct" : "incorrect"]; 
```
The `?.` is the [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#short-circuiting_evaluation).  In this case, if `content.value` is `null` or `undefined`, the expression will evaluate to `undefined`, and it won't evaluate what comes after the `.`  Otherwise, it evaluates to `content.value[correct ? "correct" : "incorrect"]`.

src/components/PathwayExercise/ExerciseContent/Assessment/styles.js
jschanker: This is very likely because the `theme` is not being passed as an input to `makeStyles` (instead it's passing in `{}`).  Do you have a route to this Component yet?  I think this problem (https://stackoverflow.com/questions/69032871/react-material-ui-themeprovider-not-being-applied) might be resolved if it becomes a descendant of the `ThemeProvider` from the [App component](https://github.com/navgurukul/bhanwari-devi/blob/815046871acd307f5492d478af6fc2c689d8d54f/src/components/App/index.js).

src/components/PathwayExercise/ExerciseContent/Assessment/index.js
Poonam-Singh-Bagh: We need to keep ordinary HTML headings as the backend team sends the variant and we have larger variant size for h1, h2, h3... in Typography

PR #419, opened 2022-06-01
Feat/volunteer automation form

Poonam-Singh-Bagh:
This is a new feature Volunteer Automation to onboard volunteers

@jschanker


src/components/VolunteerAutomation/AttendClass.js
```diff
@@ -0,0 +1,333 @@
...
+                      <Typography
+                        sx={{ mt: 2 }}
+                        gutterBottom
+                        variant="subtitle1"
+                      >
+                        {item.facilitator.name}
+                      </Typography>
+
+                      <Typography gutterBottom variant="body2">
+                        Please join at least 10 mintues before the scheduled
```
jschanker:
```suggestion
                        Please join at least 10 minutes before the scheduled
```

src/constant.js
```diff
@@ -40,8 +42,77 @@ export const HideFooter = [
...
+};
+export const lang = {
+  en: "English",
+  hi: "Hindi",
+  mr: "Marathi",
+  ta: "Tamil",
+  te: "Telugu",
+};
+export const dateTimeFormat = (date) => {
+  const datePart = date?.split("T")[0].split("-").reverse();
```
jschanker: Since the timestamp is given in `IST`, it is possible that the date of the class will be a day off for the volunteer.  For example, if the class were held at `10am on 02 Jun IST`, it should be listed as `9:30pm on 01 Jun PDT` for a volunteer on the west coast of the US.

src/components/VolunteerAutomation/AttendClass.js
```diff
@@ -0,0 +1,333 @@
...
+                          sx={{ ml: 1 }}
+                          label={languageMap[item.lang]}
+                          variant="outlined"
+                        />
+                      </Box>
+
+                      <Box sx={{ mt: 2 }}>
+                        <Typography>
+                          {/* {moment(item.start_time).format("DD-MM-YYYY")} */}
+                          {dateTimeFormat(item.start_time).finalDate} ,{" "}
```
jschanker: Is there a reason that we're defining our own function to handle `month/day/year` output formatting while using a `moment` library function for the time?

src/constant.js
```diff
@@ -40,8 +42,77 @@ export const HideFooter = [
...
+  const TimePart = date?.split("T")[1].split(":");
+  const finalDate = `${datePart[0]} ${month[datePart[1]]}, ${datePart[2]} `;
+  const finalTime = `${TimePart[0]} : ${TimePart[1]}`;
+  return { finalTime, finalDate };
+};
```
jschanker: In general, you want to try to avoid writing your own functions to manipulate `Date`s and timestamps as they can be error prone as illustrated in the above comment.  Here, you can use the built-in `Date` formatting to get something close to what you want:
```
    const finalDate = new Date(date).toLocaleString('en-GB', {month: 'short', day: "2-digit", year: "numeric"});
```
Or exactly what you want using `date-fns`:
```
    const finalDate = format(new Date(date), "dd MMM, yyyy");
``` 
Similarly, for `finalTime`, although this doesn't seem to be being used currently.
```
    const finalTime = new Date(date).toLocaleString('default', {hour: '2-digit', minute: '2-digit', hourCycle: 'h24'})
        .replace(':', ' : ');
```
Or using `date-fns`:
```
    const finalTime = format(new Date(date), "HH : mm");
```

src/constant.js
```diff
@@ -40,8 +42,77 @@ export const HideFooter = [
...
+  } else if (hours > 0) {
+    return `${hours} hrs ${minutes} mins`;
+  } else if (minutes > 10) {
+    return `${minutes} mins ${seconds} sec`;
+  } else if (minutes <= 10 && minutes > 0) {
+    return "joinNow";
+  } else {
+    return "expired";
+  }
+};
```
jschanker: Let's also try to incorporate a library function here as well.  See the `date-fns` [intervalToDuration](https://date-fns.org/v2.16.1/docs/intervalToDuration).

src/components/VolunteerAutomation/Stepper.js
```diff
@@ -0,0 +1,258 @@
...
+      },
+      (error) => {
+        console.log(error);
+      }
+    );
+  };
+
+  var today = new Date();
+  let time =
+    today.getHours() + ":" + today.getMinutes() + ":" + today.getSeconds();
```
jschanker: Is this being used anywhere?

src/constant.js
```diff
@@ -40,8 +42,77 @@ export const HideFooter = [
...
+export const dateTimeFormat = (date) => {
+  const datePart = date?.split("T")[0].split("-").reverse();
+  const TimePart = date?.split("T")[1].split(":");
+  const finalDate = `${datePart[0]} ${month[datePart[1]]}, ${datePart[2]} `;
+  const finalTime = `${TimePart[0]} : ${TimePart[1]}`;
+  return { finalTime, finalDate };
+};
+
+export const TimeLeft = (date) => {
```
jschanker: A constants file is usually used for data constants such as numbers, or strings.  Functions like `dateTimeFormat` and `TimeLeft` should probably be put in `utils.js`.  But it'd probably be better to organize these functions in a utils folder as is done here: https://github.com/google/blockly/tree/master/core/utils.  So here, we could use `date.js` to indicate date-related functionality.

src/components/VolunteerAutomation/Stepper.js
```diff
@@ -0,0 +1,258 @@
...
+import AttendClass from "./AttendClass";
+import CodeOfConduct from "./CodeOfConduct";
+import VerifyPhoneNo from "./VerifyPhoneNo";
+import IntroVideo from "./IntroVideo";
+import axios from "axios";
+import { useSelector } from "react-redux";
+import { METHODS } from "../../services/api";
+
+import "./styles.scss";
+import { parse } from "date-fns";
```
jschanker: Is this import used?

jschanker (commented):
I think we should be handling all our `Date` related functionality in a single place, specifically `utils/date.js`, instead of using different methodologies (e.g., `moment` or built-in `Date` functionality) to accomplish the same thing in different places.  It would also be good to use Date-specific library functions as much as possible to avoid having to deal with the complexities of handling different time zones, leap years, etc.  This will make it easier to migrate from one library (e.g., `moment`) to another (`date-fns`) and to address any bugs with this critical functionality.

jschanker:
Add `top/bottom margin` for one-column layout and remove the `left` margin so they line up:

image

src/components/VolunteerAutomation/Stepper.js
```diff
@@ -0,0 +1,258 @@
...
+      url: `${process.env.REACT_APP_MERAKI_URL}/volunteers/Automation`,
+      method: METHODS.POST,
+      headers: {
+        accept: "application/json",
+        Authorization: user.data.token,
+      },
+      data: {
+        contact: contact,
+        pathway_id: pathwayId,
+      },
```
jschanker, line 154: Is the `contact` and `pathwayId` stored if the potential volunteer stops filling out the form and returns later?

src/components/VolunteerAutomation/VerifyPhoneNo.js
```diff
@@ -0,0 +1,47 @@
...
+      setDisable(false);
+    }
+    if (number.length > 10) {
+      setDisable(true);
+    }
+    if (number.length < 10) {
+      setDisable(true);
+    }
+    setContact(event.target.value);
+  };
```
jschanker: `number` is reset each time this Component is re-rendered so you're essentially just writing `number = event.target.value`.  Also, why is `async` needed here?
```suggestion
  const handleChange = (event) => {
    const number = event.target.value?.replace(/[^0-9]/g, '') || '';
    setDisable(number.length !== 10);
    setContact(number);
  };
```

src/components/VolunteerAutomation/VerifyPhoneNo.js
```diff
@@ -0,0 +1,47 @@
...
+        Please let us verify your phone number
+      </Typography>
+      <Typography variant="body1" gutterBottom>
+        We will only use your number to contact you for volunteer related
+        matters. We never share it to any third party.
+      </Typography>
+      <Box sx={{ mt: 4 }}>
+        <TextField
+          label="Phone Number"
+          type="number"
```
jschanker: From the earlier discussion about why I'd recommend using `tel` instead of `number` here https://github.com/navgurukul/bhanwari-devi/pull/229#discussion_r791803501:
> Using type='tel' gives the browser (and search engines) better semantic information about the form element. If you use type = 'number', then you may have up/down arrows on the right of the form field, which could be confusing to the user as you don't usually enter telephone numbers that way

```suggestion
          type="tel"
```

jschanker (commented):
A few more comments for now, but I think we should use a maintained telephone number Component such as [react-phone-number-input](https://www.npmjs.com/package/react-phone-number-input), which would also include validation of phone numbers for volunteers from countries outside India instead whereby a couple of these comments would change.

jschanker:
Currently, the list of classes to attend isn't loading in step 4.  Also, it seems very possible to me that a volunteer might come back a number of days later after starting the signup process when they get to this step while they decide which class to attend.  So, if we store the list of classes in `localStorage`, then this list will likely contain a number of classes that already passed.  Additionally, if we're only showing a week's worth of classes now, then if the volunteer comes back at least a week later, they won't see any classes to attend.  It seems that the API call to get the list of classes should be made every time they return to step 4 and this should be the list of classes they see.  Finally, if the user isn't logged in, they will be redirected to the login page at this step.  So, should the volunteer-form be a private route?

jschanker:
Also, regarding setting items in `localStorage`, I think we might want to try to avoid using keys like `step`, which could potentially be used in another part of the app.  Instead, although slightly more inefficient, I think it might be good to create a namespace for each route like `volunteer_automation` and bundle all the state (`contact`, `step`, `pathwayId`, etc.) together.  I'll make a PR at some point in the near future to illustrate what I mean.

src/components/VolunteerAutomation/AttendClass.js
```diff
@@ -0,0 +1,333 @@
...
+  useEffect(() => {
+    dispatch(classActions.getClasses());
+  }, [dispatch]);
+
+  const languageMap = {
+    hi: "Hindi",
+    te: "Telugu",
+    en: "English",
+    ta: "Tamil",
+    cohort: "Batch",
```
jschanker: Should "Batch" be part of a `languageMap`?  If not, we can use the `lang` object constant from `src/constant.js`.

src/components/VolunteerAutomation/AttendClass.js
```diff
@@ -0,0 +1,333 @@
...
+        `${process.env.REACT_APP_MERAKI_URL}/classes/${Class.id}/register`,
+        {},
+        {
+          headers: {
+            "Content-Type": "application/json",
+            Authorization: user.data.token,
+            "register-to-all": "false",
+          },
+        }
+      )
```
jschanker, line 93: Redux again, time permitting.

src/components/VolunteerAutomation/AttendClass.js
```diff
@@ -0,0 +1,333 @@
...
+      }
+    });
+
+  const enrollClass = (Class) => {
+    setOpen(false);
+    setEnrollId(Class.id);
+    localStorage.setItem("classes", JSON.stringify([Class]));
+    setChooseClassAgain(false);
+    const notify = () => {
+      toast.success("You have been enrolled to class successfully", {
```
jschanker, line 77: Time permitting, let's start using message strings for toast notifications so that users can read text in their languages of choice.

src/constant.js
Poonam-Singh-Bagh: We don't have any outside volunteers and probably we won't have any so it won't create any issue.

src/components/VolunteerAutomation/Stepper.js
komalahire: No, we are not storing it. We will keep this in local storage.

jschanker (commented):
Few more comments here as well.  Also, should there be a link to the volunteer form in the navigation menu when the person is logged in?

src/components/VolunteerAutomation/AttendClass.js
Poonam-Singh-Bagh: No. Changed to moment.

src/constant.js
Poonam-Singh-Bagh: If we don't use this function and instead use moment then also do we have to do something?

src/constant.js
Poonam-Singh-Bagh: Sure, will do that.

src/components/VolunteerAutomation/AttendClass.js
Poonam-Singh-Bagh: No, it should not.

src/components/VolunteerAutomation/AttendClass.js
Poonam-Singh-Bagh: Ok

src/components/VolunteerAutomation/AttendClass.js
Poonam-Singh-Bagh: We didn't get this.

src/constant.js
komalahire: We are not able to use it. We have searched but are not able to implement it.
Can you help us

src/components/VolunteerAutomation/Stepper.js
Poonam-Singh-Bagh: No

src/constant.js
jschanker: @komalahire Yes, I'm going to make a PR for date-related functionality after I finalize my work on https://github.com/navgurukul/bhanwari-devi/pull/428.

src/components/VolunteerAutomation/AttendClass.js
jschanker: @Poonam-Singh-Bagh I'll make a PR request for this as well later to illustrate.

src/components/VolunteerAutomation/Stepper.js
```diff
@@ -0,0 +1,258 @@
...
+      component: <AttendClass setDisable={setDisable} />,
+    },
+    {
+      label: "Confirmation",
+      component: <Confirmation setDisable={setDisable} />,
+    },
+  ];
+
+  const isStepOptional = (step) => {
+    return step === 1;
```
jschanker: `step 1` isn't optional so since there are no optional steps as of now, you should probably use `return false` here.
```suggestion
    return false;
```

src/components/VolunteerAutomation/Stepper.js
```diff
@@ -0,0 +1,258 @@
...
+      <div
+        className="example"
+        style={{
+          overflowX: "scroll",
+        }}
+      >
+        <Stepper activeStep={activeStep}>
+          {steps.map((step, index) => {
+            const stepProps = {};
+            const labelProps = {};
```
jschanker, line 176: `labelProps` isn't being used and `stepProps` is only used when there's a skipped step, which we currently don't have.  So, these could be removed or commented out.

src/components/VolunteerAutomation/Stepper.js
```diff
@@ -0,0 +1,258 @@
...
+
+            if (isStepSkipped(index)) {
+              stepProps.completed = false;
+            }
```
jschanker, line 180: If removing/commenting out `stepProps`, this would be removed/commented out as well.  If this is the only property we'd expect to give to `stepProps`, we could also handle it in the Step property directly as:
```
completed={!isStepSkipped(index)}
```

src/components/VolunteerAutomation/Stepper.js
```diff
@@ -0,0 +1,258 @@
...
+            return (
+              <Step key={step.label} {...stepProps}>
+                <StepLabel sx={{ minWidth: "125px" }} {...labelProps}>
```
jschanker, line 183: We'd also remove `{...stepProps}` and `{...labelProps}`.

src/components/VolunteerAutomation/AttendClass.js
```diff
@@ -0,0 +1,333 @@
...
+
+  !chooseClassAgain &&
+    sliceData &&
+    sliceData.find((item) => {
+      if (item.id == enrollId) {
+        sliceData = [];
+        sliceData.push(item);
+        localStorage.setItem("classes", JSON.stringify(sliceData));
+      }
+    });
```
jschanker: No classes are shown for the user to choose to enroll in because of the following flow (will be addressing this in my PR):
1. `classes` is initially unassigned in `localStorage` so `classData` will be `null` (line 61).
2. Then `sliceData` will be an empty Array (line 63).
3. `chooseClassAgain` is `false` so `sliceData` remains an empty Array (line 64).
4. Since `sliceData` is an empty `Array`, `find` will be run on an empty Array (line 72) so `classes` will remain unassigned in `localStorage`.  Therefore when this gets re-rendered and we go back to line 61, `sliceData` remains an empty `Array`.

src/components/VolunteerAutomation/Stepper.js
```diff
@@ -0,0 +1,258 @@
...
+  const isStepOptional = (step) => {
+    return step === 1;
+  };
+
+  const isStepSkipped = (step) => {
+    return skipped.has(step);
+  };
+
+  const handleNext = () => {
+    setDisable(true);
```
jschanker: If the user goes back two steps and then goes forward one, they'll need to repeat a completed step (will be addressing this in my PR).

src/components/VolunteerAutomation/AttendClass.js
jschanker: `find` shouldn't be used to build up an Array, it should be used to get the first item in an `Array` that meets a certain condition, but it looks as if that's what you're trying to do as you only want to get one enrolled class (id)?  If so, you could do something like the following:
```
const enrolledClass = !chooseClassAgain && sliceData.find((item) => item.id === enrollId);
enrolledClass && localStorage.setItem("enrolledClass", enrolledClass); 
```
But again, we'll want to bundle the state so I'll address this in my PR.

src/components/VolunteerAutomation/AttendClass.js
```diff
@@ -0,0 +1,333 @@
...
+              textDecoration: "none",
+            }}
+            href={item.meet_link}
+            target="_blank"
+          >
+            <Button
+              onClick={() => {
+                setProceed(true);
+                localStorage.setItem("proceed", true);
+                localStorage.setItem("disabled", false);
```
jschanker: <s>Instead of keeping track of `proceed` and `disabled` separately, why don't we just check if they're enrolled in a class?</s>  **Edit**: I see that proceed essentially means attended and not trying another class.

jschanker (commented):
I'll be addressing some of these points in my PR.

jschanker:
@komalahire Is it ok if I merge #428 into this one now to avoid future conflicts caused by adding commits to this branch?  I just resolved the ones caused by the latest commit.

komalahire (approved):

PR #429, opened 2022-06-04
Design/class card

Poonam-Singh-Bagh:
Added join now button in class card new design

@jschanker


src/components/Class/ClassCard/index.js
```diff
@@ -183,60 +197,141 @@ function ClassCard({ item, editClass, enroll, style }) {
...
             <i className="check-icon check-icon fa fa-check-circle">Enrolled</i>
           )}
+          {rolesList.length == 0 && item.enrolled ? (
+            <MoreVertIcon onClick={handleOpenUserMenu} sx={{ p: 0 }} />
+          ) : (
+            rolesList.length >= 1 &&
+            (item.facilitator.email === user.data.user.email || flag) && (
+              <MoreVertIcon onClick={handleOpenUserMenu} sx={{ p: 0 }} />
+            )
+          )}
```
jschanker: The condition for which to display `MoreVertIcon` is a little complex; I feel as if it should be stored in a descriptive variable.  However, this eliminates the duplication.
```suggestion
          {(
             rolesList.length == 0 && item.enrolled ||
             rolesList.length >= 1 && (item.facilitator.email === user.data.user.email || flag)
           ) && (
             <MoreVertIcon onClick={handleOpenUserMenu} sx={{ p: 0 }} />
            )
          )}
```

src/components/Class/ClassCard/index.js
```diff
@@ -183,60 +197,141 @@ function ClassCard({ item, editClass, enroll, style }) {
...
+            style={{
+              textDecoration: "none",
+            }}
+            href={item.meet_link}
+            target="_blank"
+          >
+            <Button variant="contained" fullWidth>
+              Join Now
+            </Button>
+          </a>
```
jschanker: Can use [`ExternalLink`](https://github.com/navgurukul/bhanwari-devi/blob/08eaeab3030736341d3b7a3685c9477e3bb40de6/src/components/common/ExternalLink/index.js) component here:
```suggestion
          <ExternalLink
            style={{
              textDecoration: "none",
            }}
            href={item.meet_link}
          >
            <Button variant="contained" fullWidth>
              Join Now
            </Button>
          </ExternalLink>
```

jschanker (commented):
Couple of quick comments, will look over this more later, but some of the changes here are also duplicated in the other PR.

komalahire (approved):

src/components/Class/ClassCard/index.js
```diff
@@ -183,60 +198,137 @@ function ClassCard({ item, editClass, enroll, style }) {
       dispatch(classActions.dropOutClass(Id));
     });
   };
+
+  const EnrolledAndTimer = () => {
+    const [Timer, setTimer] = useState(TimeLeft(item.start_time));
+    const ONE_MINUTE = 60000; //millisecs
+    setInterval(() => {
+      setTimer(TimeLeft(item.start_time));
+    }, ONE_MINUTE);
```
jschanker, line 207: Few questions/points here:
1. Should this timer be being run every second instead of minute when the time gets under 1 minute or even when the time gets under 1 day?
2. By calling `setInterval` here, you will be repeatedly creating a new repeating timer with every render, which will happen at least as often as the time on the button is changed.  Instead, it'd be better to call it in a `setEffect` called once that cleans up after itself.  Something like:
```
useEffect(() => {
  const repeatingTimer =
      setInterval(() => setTimer(TimeLeft(item.start_time)), ONE_MINUTE);
  return () => clearInterval(repeatingTimer); // cleans up on unmount
}, []);
```
But if you want the time to do a variable time update (e.g., every 1 second from joining time), you can use `setTimeout` with a `[Timer]` dependency.
3. How urgent is this PR?  If it's not urgent, I can create a PR to address this and also incorporate the new function from #436 .  It shouldn't take long, but I don't know if you would need additional time to test it.

src/components/Class/ClassCard/index.js
```diff
@@ -183,60 +198,137 @@ function ClassCard({ item, editClass, enroll, style }) {
...
+            </Button>
+          </ExternalLink>
+        ) : (
+          <Button disabled={true} variant="contained">
+            Starts in {Timer}
+          </Button>
+        )}
+      </>
+    );
+  };
```
jschanker, line 228: This component is used elsewhere (e.g., volunteer automation), so I think it should be put in its own file so it can be imported by other components as needed.

jschanker, line 228: This component is used elsewhere (e.g., volunteer automation), so I think it should be put in its own file so it can be imported by other components as needed.

src/components/Class/ClassCard/index.js
```diff
@@ -183,60 +198,137 @@ function ClassCard({ item, editClass, enroll, style }) {
...
+            src={require("../assets/calendar.svg")}
+          />
+          {dateTimeFormat(item.start_time).finalDate}
         </Typography>
-        <Typography className={classes.spacing}>
-          Language : {languageMap[item.lang]}
+        <Typography variant="body1" sx={{ display: "flex" }}>
+          <img className={classes.icons} src={require("../assets/time.svg")} />
+          {moment(classStartTime).format("hh:mm a")} -{" "}
+          {moment(classEndTime).format("hh:mm a")}
```
jschanker, line 307: As before, if not urgent, we could use the `format` from the date.js file I created in my other PR that would use `date-fns` here.

jschanker, line 307: As before, if not urgent, we could use the `format` from the date.js file I created in my other PR that would use `date-fns` here.

PR #449, opened 2022-06-10
Chore/Use date-fns for dateTimeFormat

jschanker:
https://github.com/navgurukul/bhanwari-devi/pull/419#discussion_r887266941, https://github.com/navgurukul/bhanwari-devi/pull/419#discussion_r888542004, and https://github.com/navgurukul/bhanwari-devi/pull/419#discussion_r888549221

1. Replaced implementation of `dateTimeFormat` with one using `date-fns`, which is in `date.js` instead of the `constant` file.  Note: It's not necessary to do use functions like `split` and `reverse` on a date timestamp string to get the date and time.  Also, the way it was done would not account for user timezones when they differ from India's.  When `dateStr` is given as a timestamp string in the form `YYYY-MM-DDTHH:mm:ss.sssZ`, we could use `new Date(dateStr)` and can get the date and hours/minutes/seconds of the user's local time zone (See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#parameters and https://tc39.es/ecma262/#sec-date-time-string-format).  
**Here, the timestamp is in IST (`+05:30`) instead of UTC so there may be discrepancies across platforms, so this still needs to be resolved, which I'll do later.**  (Unlike `moment`, `date-fns` does not take timestamp strings with versions starting at 2.0.0. (https://blog.date-fns.org/post/we-cut-date-fns-v2-minimal-build-size-down-to-300-bytes-and-now-its-the-smallest-date-library-18f2nvh2z0yal/)
2. Added `timeRemaining` function that uses `date-fns`'s [`intervalToDuration`](https://date-fns.org/v2.16.1/docs/intervalToDuration), which has already been merged into PR #419 and some doc comments.  Again, we'll want to try to use tested functions from a `date` utility library such as date-fns to avoid the intricacies of date manipulation.

@Poonam-Singh-Bagh @kartiks26


jschanker:
https://github.com/navgurukul/bhanwari-devi/pull/419#discussion_r887266941, https://github.com/navgurukul/bhanwari-devi/pull/419#discussion_r888542004, and https://github.com/navgurukul/bhanwari-devi/pull/419#discussion_r888549221

1. Replaced implementation of `dateTimeFormat` with one using `date-fns`, which is in `date.js` instead of the `constant` file.  Note: It's not necessary to do use functions like `split` and `reverse` on a date timestamp string to get the date and time.  Also, the way it was done would not account for user timezones when they differ from India's.  When `dateStr` is given as a timestamp string in the form `YYYY-MM-DDTHH:mm:ss.sssZ`, we could use `new Date(dateStr)` and can get the date and hours/minutes/seconds of the user's local time zone (See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#parameters and https://tc39.es/ecma262/#sec-date-time-string-format).  
**Here, the timestamp is in IST (`+05:30`) instead of UTC so there may be discrepancies across platforms, so this still needs to be resolved, which I'll do later.**  (Unlike `moment`, `date-fns` does not take timestamp strings with versions starting at 2.0.0. (https://blog.date-fns.org/post/we-cut-date-fns-v2-minimal-build-size-down-to-300-bytes-and-now-its-the-smallest-date-library-18f2nvh2z0yal/)
2. Added `timeRemaining` function that uses `date-fns`'s [`intervalToDuration`](https://date-fns.org/v2.16.1/docs/intervalToDuration), which has already been merged into PR #419 and some doc comments.  Again, we'll want to try to use tested functions from a `date` utility library such as date-fns to avoid the intricacies of date manipulation.

@Poonam-Singh-Bagh @kartiks26

PR #460, opened 2022-06-12
New Class join timer button component and improved date functionality

jschanker:
#459, https://github.com/navgurukul/bhanwari-devi/pull/429#discussion_r891912328, https://github.com/navgurukul/bhanwari-devi/pull/429#discussion_r891912696, and https://github.com/navgurukul/bhanwari-devi/issues/62

* Added and improved `Date` functionality, which included converting IST timestamp to `ISOString` when passing to Date constructor for browser consistency (As per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#timestamp_string :
"When parsing date strings with the Date constructor (and `Date.parse`, they are equivalent), always make sure that the input conforms to the ISO 8601 format (`YYYY-MM-DDTHH:mm:ss.sssZ`) — the parsing behavior with other formats is implementation-defined and may not work across all browsers. Support for RFC 2822 format strings is by convention only.")
* Factored out common `ClassJoinTimerButton` from `ClassCard`/Volunteer form
* Fix: Put creation of timer in `useEffect` and cleaned up so a new one wasn't created with each passing minute and so all the repeating timers didn't remain after navigating to a different page.

@Poonam-Singh-Bagh @komalahire @kartiks26


kartiks26:
can we merge this in classBatch and then in Main?? so just to make sure nothing crashes on main site 
@jschanker

PR #463, opened 2022-06-14
Showed classes according to selected pathways

Poonam-Singh-Bagh:


komalahire (dismissed):

src/components/VolunteerAutomation/AttendClass.js
```diff
@@ -60,11 +60,21 @@ function AttendClass({
     dispatch(classActions.getClasses());
   }, [dispatch]);
 
-  const possibleClasses = data?.slice(0, numOfClassesToShow) || [];
+  const pathwayId = JSON.parse(
+    localStorage.getItem("volunteer_automation--state")
+  )?.pathwayId;
```
jschanker: If you're going to get the `pathwayId` from `localStorage`, it'd be good to use the function `getObjectState`: https://github.com/navgurukul/bhanwari-devi/blob/389625a19dc2d7760fbe168193b35e1622ca0b30/src/common/storage.js#L9-L12
So, it'd be `const pathwayId = getObjectState("volunteer_automation", "state")?.pathwayId`.  The reason for this is that these functions are responsible for selecting where it should be stored (in this case `localStorage`) and the proper `key` to use based on the provided `namespace` and `key`.

However, here I'd recommend just passing `pathwayId` as one of the `props` to `AttendClass` since it's already available in `Stepper` as we do with `enrollId`.

src/components/VolunteerAutomation/AttendClass.js
```diff
@@ -60,11 +60,21 @@ function AttendClass({
...
+
+  const classData = data?.filter((item) => item.pathway_v2 == pathwayId) || [];
+
+  const possibleClasses =
+    classData.length === 0
+      ? data?.slice(0, numOfClassesToShow) || []
+      : classData.length >= numOfClassesToShow
+      ? classData?.slice(0, numOfClassesToShow)
+      : classData;
```
jschanker: If `numOfClassesToShow` is greater than (or equal to) the `length` of `classData`, using `slice(0, numOfClassesToShow)` will simply cause the whole Array to be extracted (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice#end), so it's not necessary to do the additional check `classData.length >= numOfClassesToShow`.  Additionally, it's not necessary to use the `?` since you already set `classData` to be the empty Array when `data` is `undefined` or `null`.
```suggestion
      : classData.slice(0, numOfClassesToShow);
```

komalahire (approved):

PR #471, opened 2022-06-18
Chore/Add GitHub action to run all tests on every push/pull request

jschanker:
Allows the creation of unit tests.

1. Added workflow to build/test on every PR and on pushes to `main`, `dev`, and `classBatch` branches.
2. Deploying is commented out since this is already being done.
3. Added example test.

@kartiks26 @Poonam-Singh-Bagh @komalahire

### DIFF
--- a/.github/workflows/build_test_react.yml
+++ b/.github/workflows/build_test_react.yml
@@ -1,0 +1,45 @@
+# This is a basic workflow to help you get started with Actions
+# https://dev.to/achukka/build-and-test-react-app-with-github-actions-2jb2
+
+name: Build and Test React App
+
+# Controls when the workflow will run
+on:
+  # Triggers the workflow on push (to given branches) or pull request events
+  push:
+    branches: ["main", "dev", "classBatch"]
+  pull_request:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  build_test:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [12.x, 14.x, 15.x]
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v3
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: npm ci, build and test
+        run: |
+          npm ci
+          npm run build --if-present
+          npm test
+        env:
+          NODE_OPTIONS: "--max-old-space-size=8192"
+          CI: false
+          REACT_APP_MERAKI_URL: "https://api.merakilearn.org/"
+      # - name: deploy to gh-pages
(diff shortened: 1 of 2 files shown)


kartiks26:
@jschanker not sure how to add test and how it work . can you guide in which case we should do tests and stuff

jschanker:
@kartiks26 The GitHub action will run any test in a `.test.js` file present in a descendant of the `src` directory.  So, for example, at https://github.com/navgurukul/bhanwari-devi/runs/6997030184?check_suite_focus=true#step:4:703, you can see that the tests from `src/components/Class/ClassJoinTimerButton/ClassJoinTimerButton.test.js` were run and passed successfully.

In general, you want to write tests for every `Component` and one of the conventions for doing so would be to make the name of the file the same as the name of the `Component`, adding the `.test.js` at the end.  Another option would be to put all the tests together in a `tests` directory.

Related tests can be grouped together with `[describe]` (https://jestjs.io/docs/api#describename-fn) as we could do with grouping together tests for a given Component:
https://github.com/navgurukul/bhanwari-devi/blob/6643a81cdf33cf5bbeaa5715dedd56bae2efb941/src/components/Class/ClassJoinTimerButton/ClassJoinTimerButton.test.js#L8

Inside a `describe`, you'd create tests which you could do with `test` or the alias `it` (https://jestjs.io/docs/api#testname-fn-timeout): https://github.com/navgurukul/bhanwari-devi/blob/6643a81cdf33cf5bbeaa5715dedd56bae2efb941/src/components/Class/ClassJoinTimerButton/ClassJoinTimerButton.test.js#L35-L36

These `test`s should be checking that the requirements of this particular `Component` are met (assuming anything it relies on work correctly) and should be somewhat stable in that changes to an implementation detail in the future shouldn't make them fail.   Ideally, tests should also be thoroughly making sure every code path works correctly.  So, if you have an `if/else if/else`, the tests should ensure that the code in each branch works correctly.  In short, unit tests should be checking automatically what you'd be checking yourself if you didn't have them available.  In [test-driven development](https://en.wikipedia.org/wiki/Test-driven_development), you actually write the tests before you finish writing the code so the tests actually drive the development.

For each test, you have a set of expectations that provide evidence that what you're testing for actually works.  This is done using `expect` (https://jestjs.io/docs/expect):  
https://github.com/navgurukul/bhanwari-devi/blob/6643a81cdf33cf5bbeaa5715dedd56bae2efb941/src/components/Class/ClassJoinTimerButton/ClassJoinTimerButton.test.js#L29-L31

A test passes exactly when all expectations called inside the function provided to it are `true`.  The above was actually used in two different cases, one to check if the lone button that renders as a result of the component is enabled and one to check if it's disabled:
```
expect(getByRole("button")).toBeEnabled();
expect(getByRole("button")).toBeDisabled();
```
For a case where the button is enabled, `expect(getByRole("button")).toBeEnabled();` will be `true` whereas it'll be `false` if the button is disabled.  As mentioned, a failed expectation will result in a failed test.  

There are a number of simpler matchers you could use such as `expect(_____).toBeFalsy()` (https://jestjs.io/docs/expect#tobefalsy) `expect(_____).toBeGreaterThan()` (https://jestjs.io/docs/expect#tobegreaterthannumber--bigint) that you can read about here: https://jestjs.io/docs/expect.

PR #479, opened 2022-06-22
Added the video in the pathwaypage

amansharmma:
Added the video as a link at (What's it all about?)
Screenshot from 2022-06-22 13-41-59


src/components/PathwayCourse/index.js
```diff
@@ -24,10 +24,12 @@ import {
   Typography,
   CardMedia,
   Button,
+  Se,
```
jschanker, line 27: Added accidentally?

src/components/PathwayCourse/index.js
```diff
@@ -214,6 +221,35 @@ function PathwayCourse() {
                     <Typography variant="body1">
                       {pathwayCourseData.description}
                     </Typography>
+
+                    <ExternalLink
+                      style={{
+                        textDecoration: "none",
+                      }}
+                      href={"https://youtu.be/DDFvJmC3J5M"}
```
jschanker, line 229: You can remove the `{` and `}` when using a string literal as you're doing here.

src/components/PathwayCourse/index.js
```diff
@@ -152,6 +154,11 @@ function PathwayCourse() {
       "Here"
     );
   }, [upcomingBatchesData, userEnrolledClasses]);
+
+  const handleVideo = () => {
+    return <YouTube className={"youtube-video"} videoId="DDFvJmC3J5M" />;
```
jschanker, line 159: Same here, braces shouldn't be needed for a string literal `className` attribute value.

src/components/PathwayCourse/index.js
```diff
@@ -214,6 +221,35 @@ function PathwayCourse() {
...
+                        align="start"
+                        variant="body2"
+                      >
+                        <img
+                          src={require("./asset/ComputerScreen.svg")}
+                          alt="MonitorScreen Img"
+                        />
+                        <section
+                          className={classes.link}
+                          onClick={handleVideo}
```
jschanker, line 243: This is not needed as the function just returns a React element, which is not used.  If you wanted to embed the YouTube video instead of link to it, you might use `{handleVideo()}` on line 250, but then you'd also rename the function like `IntroVideo` to indicate it's a component and not a click handler.

src/components/PathwayCourse/index.js
```diff
@@ -152,6 +154,11 @@ function PathwayCourse() {
       "Here"
     );
   }, [upcomingBatchesData, userEnrolledClasses]);
+
+  const handleVideo = () => {
+    return <YouTube className={"youtube-video"} videoId="DDFvJmC3J5M" />;
+  };
```
jschanker, line 160: This could be removed since it won't be used now.

src/components/PathwayCourse/index.js
```diff
@@ -214,6 +221,35 @@ function PathwayCourse() {
...
+                        <section
+                          className={classes.link}
+                          onClick={handleVideo}
+                          style={{
+                            cursor: "pointer",
+                          }}
+                        >
+                          {"  "} &nbsp; &nbsp;
+                          <b>What's it all about?</b>
+                        </section>
```
jschanker, line 250: Avoid use a `section` for link text.  See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section for when it should be used.

src/components/PathwayCourse/index.js
```diff
@@ -214,6 +221,35 @@ function PathwayCourse() {
...
+                        />
+                        <section
+                          className={classes.link}
+                          onClick={handleVideo}
+                          style={{
+                            cursor: "pointer",
+                          }}
+                        >
+                          {"  "} &nbsp; &nbsp;
+                          <b>What's it all about?</b>
```
jschanker, line 249: Avoid using `<b>` if the only purpose is to make the text bold.  You can use CSS for this.  See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b for when you should use `<b>`.

src/components/PathwayCourse/styles.js
```diff
@@ -72,9 +72,11 @@ const useStyles = makeStyles((theme) => ({
     paddingTop: "16px",
   },
   certificateLogo: {
-    // marginBottom : "5px",
     fontSize: "90px",
   },
+  link: {
+    color: "green",
```
jschanker, line 78: Use theme colors for consistency in styles and to allow for future changes to be made in one place.

PR #489, opened 2022-06-26
Feat/framework for language support

jschanker:
Adds a framework for multilingual support.  For example, a language constant such as `MSG.LEARN` can be used in the header so that when `?hl=en` is put at the end of the URL, this will appear as the English `Learn`, but for `?hl=hi`, this will appear in Hindi as सीखें (`hl` stands for host language and can be used for Google, e.g., https://www.google.com/?hl=hi).

English (en):
image

Hindi (hi):
image

https://github.com/navgurukul/bhanwari-devi/pull/419#discussion_r889494283

**How to add message constants in a Component**
1. Add the following `import` statements in the file with the Component (possibly adding `../` if the Component is in a subdirectory):
```
import { useContext } from "react";
import { LanguageProvider } from "../../common/context";
```
2. Get the object `MSG` with the current language constants (`language` is used below as well, but is not necessary if you don't want to also get the code of the current language such as `hi` or `en`) by adding the following in the React component:
```
const { language, MSG } = useContext(LanguageProvider); 
```
3. If the constant name is `LOADING`, use `MSG.LOADING` in the Component.
4. Add `en.LOADING = "Loading..."; ` to `src/msg/en.js`.  If you want to add support for Hindi, add `hi.LOADING = "खुल रहा है"` to `src/msg/hi.js`.
5. (Optional) For other languages such as Marathi, create a new file in `src/msg` with that code (`mr.js`) and add the following:
```
const mr = {};
mr.LOADING = "(translation for loading in Marathi)";

export default mr;
```
6. Then add an import (or uncomment the existing one) and add it to `MSG` in `src/msg/index.js`:
```
import mr from "./mr.js";
const MSG = { en, hi, mr }; //, ta, te};
```

1. Added a `LanguageProvider` context in `src/common/context.js`
2. Added a `useQuery` custom hook from https://v5.reactrouter.com/web/example/query-parameters in a new file `src/common/url.js` that can extract values from their corresponding names in the query string. 
3. Added a `useLanguage` custom hook that uses `useQuery` to get the value of `hl` if it's present or otherwise gets it from `localStorage` if it's present.  If it's a valid language code, it stores it in `localStorage` and returns that code.  Otherwise, it returns `en` for English.  Also added a helper function `getTranslationKey` to get the key of an English value, if it exists in `en.js` (e.g., `getTranslation("Get Involved")` would evaluate to `GET_INVOLVED`). All this was added in a new file `src/common/language.js`.
4. Added a `LanguageContext` Provider in `App.js` that provides the current language (returned from `useLanguage`) along with the constants from the language.
5. Added `en.js` and `hi.js` which export language constant objects for Hindi and English.
6. Added `msg/index.js` which exports an object with all the languages from `lang` as keys and the objects exported from the various language files as values augmented by the values from English for any key they don't possess (i.e., for which no transalation has been provided).  
7. Use the `MSG` property from the `LanguageContext` to get the translations of "Learn", "About" and "Get Involved" in the header/dropdown header based on the selected language.

@Poonam-Singh-Bagh @komalahire @kartiks26


jschanker:
Student view navigation:
image
For 4 roles:
image
For 2 Roles:
image
Search:
image

PR #431, opened 2022-06-06
Class And Batch UI for Review

kartiks26:
Class and batch Ui is part where user will enroll in a batch which contains 28 classes pre enrolled and every exercise / pathway course will have random,classes for same generated by respective teacher .
Upcomming batches will shown only on login and exercises are also now on will be logged in Based.
In this PR the most of work is on Pathway Course .

1. /pathway/1
image
image
image
image

2. /course-content/1/370/1 
image
image

3. 
image

4. /profile
image

5. Popups 
image
image

6. user Dashboard
image

7. Work in progress for LEarning track which is also part of this update.

Note : in Exercise Content Might Crash As work progress on Learning Track 

@jschanker

PR #524, opened 2022-07-19
Fix/Add / after base URL environment variables

jschanker:
https://github.com/navgurukul/bhanwari-devi/pull/431#discussion_r891339459 : 
> IMPORTANT: There are a number of places that do not contain a trailing / after ${process.env.REACT_APP_MERAKI_URL}. From what I can tell, this will refer to https://dev-api.merakilearn.org/ in the development environment, but https://api.merakilearn.org (without the trailing /) in production. So, this would refer to say https://api.merakilearn.orgpathways/1/upcomingBatches, causing bad requests and lots of problems after merging into main (You can see these requests if you open up the Developer console and check the Network tab before visiting https://pr-431.dkchei85ij0cu.amplifyapp.com/course-content/1/370/1).

* Performed a recursive search of all files in `src` folder repository for regex `{process.env.[^}]*}[^/]` and added `/` at end where appropriate.

**Action items**
Environment variables that should have their trailing slash removed, if present:
```
REACT_APP_MERAKI_URL
REACT_APP_ADMISSIONS_URL
REACT_APP_CHANAKYA_BASE_URL
```

@kartiks26 @anandpatel504

### DIFF
--- a/src/components/BatchClassComponents/Revision/RevisionClassEnroll.js
+++ b/src/components/BatchClassComponents/Revision/RevisionClassEnroll.js
@@ -30,7 +30,7 @@ function RevisionClassEnroll(props) {
   useEffect(() => {
     axios({
       method: METHODS.GET,
-      url: `${process.env.REACT_APP_MERAKI_URL}classes/${id}/revision`,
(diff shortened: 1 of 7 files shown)


kartiks26:
@jschanker  i alerady did this in branch mereged into main
so will close this for now

PR #536, opened 2022-07-26
Perf/GH action/script to combine back- and front-end pathway data in static file

jschanker:
Adds GH action to make API call to get and store back-end pathway data, and make a call to a new script to join this pathway data with a static file of front-end pathway data and write out the combined output to the repository.  Run on every push and approximately every hour, depending on the reliability of GH action `cron` jobs.  API call to get pathway data is replaced with request for static file with combined pathway data, saving many API requests to the server.

**Action items**
1. Currently retrieving back-end pathway data from https://dev-api.merakilearn.org.  Action should be updated to refer to https://api.merakilearn.org before merging into main.
2. Front-end data will need to be updated in static file [/src/data/front-end_pathway_data_v40.json](https://github.com/navgurukul/bhanwari-devi/blob/gh-action-store-combined-fe-and-be-pathway-data/src/data/front-end_pathway_data_v40.json) as appropriate.
3. Paths for Misc. courses and Residential courses are hard-coded in GH action so this might need to be changed or maybe it shouldn't be part of the front-end data file.
4. Not sure whether link to raw data file will break when this PR is merged, so may need to update the link to the corresponding raw data file in the dev branch if the link breaks.  

@kartiks26


kartiks26:
so Should we merege this to dev or have a complete solution and then merge

kartiks26:
@anandpatel504 Give your thoughts about this

kartiks26:
@jschanker before merging this to main what will need to do or we can use this right away like this

jschanker:
@kartiks26 The action items above would need to be addressed:  
1. Currently retrieving back-end pathway data from `https://dev-api.merakilearn.org/`. Action should be updated to refer to `https://api.merakilearn.org/` but ideally, we should be using an environment variable referring to a secret (`https://api.merakilearn.org/`) for this as I do here: https://github.com/jschanker/githubactions-testing/blob/main/.github/workflows/batch-data.yml#L12.  We'd probably want to do this with other environmental variables referring to file paths in the workflow, although this is not strictly necessary.
2. Front-end data such as `video_link` will need to be updated in static file [/src/data/front-end_pathway_data_v40.json](https://github.com/navgurukul/bhanwari-devi/blob/gh-action-store-combined-fe-and-be-pathway-data/src/data/front-end_pathway_data_v40.json) as appropriate, if we're going to use it.
3. Paths for Misc. courses and Residential courses are hard-coded in GH action so this might need to be changed, again if we're going to use it, or maybe it shouldn't be part of the front-end data file.
4. The raw data file is still present now that this PR is merged, but we'd want the `axios` request to be made to what's now on the `dev` branch (or maybe the `main` one), so it'll continue to get updated (https://raw.githubusercontent.com/navgurukul/bhanwari-devi/dev/src/data/pathway_data_v40.json) since the merged branch has been deleted.

Also, we're currently not making use of the combined front-end data as it's still being hard-coded (e.g., in https://github.com/navgurukul/bhanwari-devi/blob/dev/src/components/Footer/index.js#L16-L32).  However, it's fine to just ignore the combined front-end data for now.  One other thing we might want to do is to `catch` any errors with the `axios` request for the static file with the ordinary API request and joining the data as a backup in case something goes wrong with accessing the file: https://github.com/navgurukul/bhanwari-devi/blob/273a299c74b953b77dbe5a4fb0e2c2c8c84f02b2/src/components/PathwayCourse/redux/api.js#L8-L53)

PR #551, opened 2022-07-30
Feat/new class form

Poonam-Singh-Bagh:
1. Created new design class form 
2. Created 2 Components `ClassForm` and `SuccessModal`
3. SuccessModal will be calling in `ClassForm` when a class gets created successfully
4. Some functions in ClassForm
    A. `createClass` and `editClass` for creating and editing the class respectively
    B. `convertToIST` to change the date formate
    C. `submitHandle` to handle the submission and to decide object's which property is for doubt class and batch
5. There is a functionality handling in `submitHandle` to change the date according to on_days

Use @vivekdogra @jschanker  key to tag people to review from


src/components/Class/ClassForm.js
```diff
@@ -0,0 +1,720 @@
...
+    );
+    const utc = dateInObj.getTime() + dateInObj.getTimezoneOffset() * 60000;
+    return new Date(utc + 3600000 * +5.5).toISOString();
+  };
+
+  const createClass = (payload) => {
+    payload.start_time = convertToIST(payload.start_time);
+    payload.end_time = convertToIST(payload.end_time);
+    return axios({
+      url: `${process.env.REACT_APP_MERAKI_URL}classes`,
```
jschanker: Remember that we'll need a slash in the path to address https://github.com/navgurukul/bhanwari-devi/pull/431#discussion_r891339459 :

> IMPORTANT: There are a number of places that do not contain a trailing / after ${process.env.REACT_APP_MERAKI_URL}. From what I can tell, this will refer to https://dev-api.merakilearn.org/ in the development environment, but https://api.merakilearn.org (without the trailing /) in production. So, this would refer to say https://api.merakilearn.orgpathways/1/upcomingBatches, causing bad requests and lots of problems after merging into main (You can see these requests if you open up the Developer console and check the Network tab before visiting https://pr-431.dkchei85ij0cu.amplifyapp.com/course-content/1/370/1).
```suggestion
      url: `${process.env.REACT_APP_MERAKI_URL}/classes`,
```

src/components/Class/ClassForm.js
```diff
@@ -0,0 +1,720 @@
...
+    const dateInObj = new Date(
+      Date.UTC(b[0], --b[1], b[2], b[3], b[4], b[5], b[6])
+    );
+    const utc = dateInObj.getTime() + dateInObj.getTimezoneOffset() * 60000;
+    return new Date(utc + 3600000 * +5.5).toISOString();
+  };
+
+  const createClass = (payload) => {
+    payload.start_time = convertToIST(payload.start_time);
+    payload.end_time = convertToIST(payload.end_time);
```
jschanker, line 193: Is it necessary to convert to IST here @vivekdogra @anandpatel504 ?  At quick glance of what's happening in the back-end, it seems this is being done here anyway: https://github.com/navgurukul/sansaar/blob/main/lib/routes/classes.js#L322-L325 or https://github.com/navgurukul/sansaar/blob/main/lib/routes/classes.js#L358-L359.  I'm actually not sure why we ever need to convert to IST (back-end or front-end) and aren't just storing/handling everything in `UTC` ?  Regardless, I recommend all date-related functionality be handled centrally: https://github.com/navgurukul/bhanwari-devi/blob/dev/src/common/date.js#L219-L237 .

src/components/Class/ClassForm.js
```diff
@@ -0,0 +1,720 @@
...
+      },
+      (error) => {
+        console.log("error", error);
+      }
+    );
+  };
+
+  const editClass = (payload) => {
+    payload.start_time = convertToIST(payload.start_time);
+    payload.end_time = convertToIST(payload.end_time);
```
jschanker, line 226: Same comment about conversion to IST here

src/components/Class/ClassForm.js
```diff
@@ -0,0 +1,720 @@
...
+        fieldsToSend.max_enrolment = classFields.max_enrolment;
+      }
+      createClass(fieldsToSend);
+    } else {
+      if (classFields.max_enrolment != "No Limit") {
+        //add max_enrolment field only if it is not No Limit
+        fieldsToSend.max_enrolment = classFields.max_enrolment;
+      }
+      editClass(fieldsToSend);
+    }
```
jschanker: Use the `if/else` only for what's different between the cases:
```suggestion
    if (classFields.max_enrolment != "No Limit") {
      //add max_enrolment field only if it is not No Limit
      fieldsToSend.max_enrolment = classFields.max_enrolment;
    }
    (!isEditMode ? createClass : editClass)(fieldsToSend);
```

src/components/Class/ClassForm.js
```diff
@@ -0,0 +1,720 @@
...
+            <Grid container mb={4}>
+              <Grid item xs={11}>
+                <Typography variant="h6" component="h2">
+                  {isEditMode
+                    ? `Update ${
+                        classFields.type == "batch" ? "Batch" : "Doubt Class"
+                      }`
+                    : `Create ${
+                        classFields.type == "batch" ? "Batch" : "Doubt Class"
+                      }`}
```
jschanker:
```suggestion
                  {
                    (isEditMode ? 'Update ' : 'Create ') +
                      (classFields.type == 'batch' ? 'Batch' : 'Doubt Class');
                  }
```

src/components/Class/ClassForm.js
```diff
@@ -0,0 +1,720 @@
...
+        while (i <= 7) {
+          incrementedDate = moment(incrementedDate).add(1, "days")._d;
+          let Day = incrementedDate.toString().split(" ")[0];
+          if (days[weekDday[index]] === Day) {
+            newDate = incrementedDate;
+            break;
+          }
+          i = i + 1;
+        }
+        classFields.date = moment.utc(newDate).format("YYYY-MM-DD");
```
jschanker: One of a number of places where I'd recommend replacing references to `moment`'s `format` with the `format` from [date.js](https://github.com/navgurukul/bhanwari-devi/blob/dev/src/common/date.js#L239-L257) which uses `date-fns` instead of the legacy `moment`.  Note that the argument to format will changes to `yyyy-MM-dd` (as per https://github.com/date-fns/date-fns/blob/main/docs/unicodeTokens.md#popular-mistakes).  As with an earlier comment, I'll need to look more closely into these conversions between UTC and IST.

jschanker (commented):
Some initial comments

src/components/Class/ClassForm.js
```diff
@@ -0,0 +1,720 @@
...
+                        setClassFields({
+                          ...classFields,
+                          ["end_time"]: endTime,
+                        });
+                      }}
+                      renderInput={(params) => <TextField {...params} />}
+                    />
+                  </Stack>
+                </LocalizationProvider>
+              </Grid>
```
jschanker: Not sure that it matters too much here, but maybe remove the duplication:
```suggestion
              {[
                { label: 'Start Time', prop: 'start_time' },
                { label: 'End Time', prop: 'end_time' },
              ].map(({label, prop}) => (
                <Grid item xs={isActive ? 12 : 6}>
                  <LocalizationProvider dateAdapter={AdapterDateFns}>
                    <Stack spacing={3}>
                      <DesktopTimePicker
                        label={label}
                        value={classFields[prop]}
                        onChange={(time) => {
                          setClassFields({
                            ...classFields,
                            [prop]: time,
                          });
                        }}
                        renderInput={(params) => <TextField {...params} />}
                      />
                    </Stack>
                  </LocalizationProvider>
                </Grid>
              ))}
```

src/components/Class/ClassForm.js
```diff
@@ -0,0 +1,720 @@
...
+        end_time: classFields.end_time,
+        partner_id: classFields.partner_id,
+        category_id: classFields.category_id,
+        pathway_id: classFields.pathway_id,
+        lang: classFields.lang,
+        frequency: classFields.frequency,
+        type: classFields.type,
+        on_days: classFields.on_days,
+      };
+    }
```
jschanker: You can use Lodash's [pick](https://lodash.com/docs/4.17.15#pick) to make this code a little less verbose.  Also, avoid using `var` as there is almost never a reason to use this over `let` or `const` now:
```suggestion
    const commonFields = [
      "title",
      "description",
      "start_time",
      "end_time",
      "category_id",
      "pathway_id",
      "lang",
      "type",
    ];
    let fieldsToSend;
    if (classFields.type === "doubt_class") {
      fieldsToSend = _.pick(classFields, [
        ...commonFields,
        "course_id",
        "exercise_id",
      ]);
    } else if (classFields.type === "batch") {
      fieldsToSend = _.pick(classFields, [
        ...commonFields,
        "partner_id",
        "frequency",
        "on_days",
      ]);
    }
```

src/components/Class/ClassForm.js
Poonam-Singh-Bagh: Ohh, yes... I had forgot, thanks.

src/components/Class/ClassForm.js
Poonam-Singh-Bagh: @jschanker What should I do then? Need to change `moment`'s `format` or should I change it to `date-fns`?

jschanker: @Poonam-Singh-Bagh You'll want to use `date-fns`, but not directly.  Since you want to convert to UTC before formatting, you could use the commented out [formatInUtc](https://github.com/navgurukul/bhanwari-devi/blob/main/src/common/date.js#L235-L237) as `formatInUtc(newDate, "yyyy-MM-dd")`.  You also need to uncomment out the [formatInTimezone](https://github.com/navgurukul/bhanwari-devi/blob/main/src/common/date.js#L231-L233) and the import statement at the [top](https://github.com/navgurukul/bhanwari-devi/blob/main/src/common/date.js#L9) (https://github.com/navgurukul/bhanwari-devi/blob/main/src/common/date.js#L9)

In `moment`, `dd` will refer to the [weekday](https://momentjs.com/docs/#/displaying/format/:~:text=1st%20...%205th%206th-,dd,Su%20Mo%20...%20Fr%20Sa,-ddd).

```suggestion
        classFields.date = formatInUtc(newDate, "yyyy-MM-dd");
```

PR #553, opened 2022-08-01
Perf/GitHub action/script to store student class and upcoming batch data

jschanker:
GitHub action/script to store and commit all student classes from the past week and then extract/store the upcoming student batches.  Makes request to static file instead of API endpoint to get upcoming batches for students to decrease the number of API requests and loading time for the student. 

1. Changed implementation to get upcoming batches to (1) consider it an upcoming batch if the first class already started but didn't yet complete (i.e., the time is after the first class's start time but before its end) and (2) only go through the Array of classes once by using a `Map` to store the time of the last class for that given recurring id.  Although this shouldn't matter given the small number of batches, it ensures `O(n)` or `O(nlgn)` runtime (depending on whether looking for a key/setting the value of a key in the Map is an `O(1)` or `O(lg n)` operation) where `n` is the number of classes in a 2-week period. 

**Action items**
1. Need to store student API token in secrets for this to work.
2. Need to change https://dev-api.merakilearn.org to https://api.merakilearn.org when merging into production to get the correct class data.


kartiks26:
will this be real-time as a volunteer creates a batch will it reflect to student as soon as it is created??

jschanker:
> will this be real-time as a volunteer creates a batch will it reflect to student as soon as it is created??

@kartiks26 Not currently, but we can schedule it to run to update the class/batch data when that happens.

kartiks26:
this calls are related to partner ID of every candidate so need realtime and cant be cashed as static file i guess
or we can cash all latest and then map only those which maps the partner iD of that student

jschanker:
With respect to getting all future classes, I haven't looked into an API endpoint I could use that does this.  The `classes` one currently only gets classes from now until a week later.  But the `classes/all` one only gets those matching the partner id of the student's so we would need to make multiple requests (and add multiple tokens in secrets).  @anandpatel504 maybe we can create a new API endpoint for getting all classes starting from a certain date, not filtered by matching `partner_id` to facilitate this?  The decrease in load time is substantial so seems worth it.  We do need to make sure the file isn't cached by the browser though if we want to allow for updates (same with pathways).

**Edit**: Disregard about creating a new endpoint as I realized the `admin` role will get all the classes from every partner id (`null` or not).

PR #683, opened 2022-09-12
OTP VOULUNTEER FLOW AUTH 

kartiks26:
> Supports International numbers
> add country code at start
> Snackbar added
UI 
image


src/components/VolunteerAutomation/VerifyPhoneNo.js
```diff
@@ -1,14 +1,117 @@
...
+const app = initializeApp(firebaseConfig);
 
 function VerifyPhoneNo({ setDisable, setContact, contact }) {
   const handleChange = (event) => {
     const number = event.target.value?.replace(/[^0-9]/g, "") || "";
-    setDisable(number.length !== 10);
-    setContact(number);
+    // give space only after first two digits
+    if (number.length > 2) {
+      const formattedNumber = `${number.slice(0, 2)} ${number.slice(
```
jschanker, line 37: Country codes won't always be 2 digits. For example, for the US, it's `+1` and for Finland, it's `+358`. Do we want to go back to using the `MuiPhoneNumber` component from `material-ui-phone-number` instead of using an ordinary text input and manually applying formatting (https://github.com/navgurukul/bhanwari-devi/pull/445)? Also, we can use `parseMax` from `libphonenumber-js` (https://github.com/navgurukul/bhanwari-devi/pull/445/files#diff-ac8c0858255c4d048a7d55314cf472fbf716e242bc8c2798abcc30e0828127eeR4) for basic validation (land line, mobile, etc.) before using up verification codes. One downside of this validation is increasing app bundle size.

kartiks26: @jschanker yes we can take this up and do it

kartiks26: Thankyou .

kartiks26: For now merging this to dev and will give task to @amansharmma

PR #758, opened 2022-10-16
refactor Assessment code

### LINKED ISSUE #757: Refactor code for Assesment
The current code written in assessment has some Redundant states that can be eliminated and code can be made more clear.
Time take --> 2 Days (17/10)

kartiks26:
Which issue does this refer to ?
 #757 

image

### DIFF
--- a/src/components/PathwayExercise/ExerciseContent/Assessment/AssessmentContent.js
+++ b/src/components/PathwayExercise/ExerciseContent/Assessment/AssessmentContent.js
@@ -1,0 +1,206 @@
+import React from "react";
+import { Grid, Typography, Box, Button, Paper, Stack } from "@mui/material";
+import useStyles from "../../styles";
+import get from "lodash/get";
+import DOMPurify from "dompurify";
+function UnsafeHTML(props) {
+  const { html, Container, ...otherProps } = props;
+  const sanitizedHTML = DOMPurify.sanitize(html);
+  return (
+    <Container
+      {...otherProps}
+      dangerouslySetInnerHTML={{ __html: sanitizedHTML }}
+    />
+  );
+}
+
+const headingVarients = {};
+
+[Typography, "h2", "h3", "h4", "h5", "h6"].forEach(
+  (Name, index) =>
+    (headingVarients[index + 1] = (data) => (
+      <UnsafeHTML
+        Container={Name}
+        className="heading"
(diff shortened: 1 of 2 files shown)


kartiks26:
@saquibaijaz Can you review or test it??
https://meraki-dev-git-refactor-assesment-sdevmeraki.vercel.app/

saquibaijaz:
@kartiks26 
What do I need to check here? There is no description of the issue.

kartiks26:
Refractor some code, can you check is assessment working fine?
@saquibaijaz

saquibaijaz:
@kartiks26 
1. The study material does not get "Marked as done" if we click NEXT for the first time. We either need to click NEXT again, click BACK, or close the module and then go to the same module; then only it is marked as done
2. The question, if answered correctly on the first attempt or after retry, or even if we see the explanation does not get the question "Marked as done", we need to click NEXT or BACK or go out of the course module and then come back, then only it is "Marked as done".

The expected behaviour is:
1. For study material: If NEXT is clicked, it should get marked as done
2. For questions: Will be marked as done only when:
- the question is answered correctly on the first attempt
- the question is answered correct or wrong in the second attempt
- see explanation is clicked
- (in short, we can say that whenever the explanation is displayed, the question will be marked as done)

saquibaijaz:
@kartiks26 Still not working

https://user-images.githubusercontent.com/111858763/196952382-da26b673-789a-4d5c-be50-7cf8da848d0d.mp4


https://user-images.githubusercontent.com/111858763/196952408-dd316fa9-a34f-4f2d-9e0c-122fe13fa10d.mp4

kartiks26:
@saquibaijaz I mentioned this right !
Its not under the asessment part . its regarding to progress tracking.
And that progress tracking stuff will require redux saga support which @Poonam-Singh-Bagh is good of.
And was planning to convert it into redux toolkit but that requires the upgrade of node version which @destro1108 was working on.
So its depending on either of poonam or swanand who can help me along with any way.

saquibaijaz:
@kartiks26 All right. But I was told to check whether the assessment was working correctly or not. 
If that doesn't need to be checked here, then I think I would be unable to check the refactored code. 
You guys can give a sign-off for this, and it can be moved to prod.

kartiks26:
Yes sure!!

PR #760, opened 2022-10-17
added Supplemental english course

komalahire:


src/components/PathwayCourse/index.js
```diff
@@ -191,6 +190,20 @@ function PathwayCourse() {
     return item.id == pathwayId;
   });
 
+  let SupplementalCourse;
+  let filterPathwayCourse;
+
+  if (pathwayId == 2) {
+    filterPathwayCourse = pathwayCourse?.data?.courses.filter(
+      (item) => item.name === "Spoken-English"
```
kartiks26: Here as well if object item don`t contained name in it app might crash `item?.name` will be better to be used

src/components/PathwayCourse/index.js
```diff
@@ -191,6 +190,20 @@ function PathwayCourse() {
...
+    );
+
+    SupplementalCourse = pathwayCourse?.data?.courses.filter(
+      (item) => item.name !== "Spoken-English"
```
kartiks26: same as last

kartiks26 (commented):
@komalahire Added some comments regarding the Destructuring

PR #768, opened 2022-10-25
chng in name

saloni-080601:
A clear and concise description. Use `#` to refer to github issues and PR

1. Upgraded packages
3. ...


src/components/Tutor/BaseURL.js
```diff
@@ -0,0 +1,68 @@
...
+  URL = `${process.env.REACT_APP_MERAKI_URL}/volunteers`;
+
+  /*Setting date range */
+  setDates(fromStart, toEnd) {
+    this.URL = this.URL.concat(`?from=${fromStart}&to=${toEnd}`);
+  }
+
+  /*Filters and Search */
+  /*  Total varialbles = n = 3
+        Number of cases = nC3 + nC2 + nC1 = 1 + 3 + 3 = 7
```
jschanker: Probably easier to get the number of possibilities by thinking `n` filters, `2` choices for each, which would give you `2**n`. But since you're excluding the no filter choice, there's one less choice so you're using `2**n - 1`.

src/components/Tutor/BaseURL.js
```diff
@@ -0,0 +1,68 @@
...
+      /*Only search present*/
+      case isValid(searchTerm):
+        this.URL = this.URL.concat(`?name=${searchTerm}`);
+        break;
+      /*Only language present*/
+      case isValid(langFilter):
+        this.URL = this.URL.concat(`?lang=${langFilter}`);
+        break;
+    }
+  }
```
jschanker: What if you want to add more filters in the future? Every time you add 1, you're going to be doubling the number of possibilities. This will mean more inputs to `statusFilter` and exponentially growing duplication of code. It'd be better to use an object input with keys and values (e.g., `{status: statusFilter, lang: langFilter, name: searchTerm}`). But for now, you could still remove the code duplication:
```suggestion
    const filters = {status: statusFilter, lang: langFilter, name: searchTerm};
    this.URL += "?" + Object.entries(filters)
      .filter((keyValuePair) => isValid(keyValuePair[1]))
      .map((keyValuePair) => `${keyValuePair[0]}=${keyValuePair[1]}`)
      .join("&");
```
Another possibility is to use `toString` on a `URLSearchParams` object, but you'd still need to `filter` out the invalid properties first.

src/components/Tutor/BaseURL.js
```diff
@@ -0,0 +1,68 @@
+function isValid(val) {
+  return val !== "";
+}
```
jschanker: For filters, you might want to have a different `isValid` function for each one.

PR #785, opened 2022-11-03
Certificate Generation

### LINKED ISSUE #755: Certificate Generation
### Description of the Problem

The certificates are issued manually at the moment. We need to run a script with the student details after every batch completion. It is cumbersome to keep track.

### Describe the solution you'd like

The solution we want is to automate certificate generation based on the student progress from the app or website itself. Checklist for the current solution is as follows:

- [x] The student will see a greyed out certificate icon in the course tree after the last course i.e. Functions.
- [x] When the student clicks on the greyed out certificate icon, a toast message is displayed urging them to complete all courses to unlock the certificate.
- [x] When the students has attended all the classes and completed all the reading material and MCQs, the certificate will be unlocked.
- [x] Student will be able to see a preview of the certificate.
- [x] Able to download the certificate as a PDF.
- [ ] Able to share the certificate externally or in Meraki chat (this may come later).

### Pendency on
The progress should work seamlessly across the platforms in order for us to determine the student's progress and issue them certificates.

### Future Plans

A mid-phase certificate may be brought at a later stage. Also, there will be milestones for completing each course. However, they are different from certificates.

itsomkathe:
Refer to issue #755 

@komalahire @jschanker


src/components/PathwayCourse/index.js
```diff
@@ -113,6 +136,28 @@ function PathwayCourse() {
...
 
+  const handleSnackbar = ()=>{
+    setOpenSnackbar(prev=>!prev);
+  }
+
+  const handleModal = ()=>{
+    if(!openModal){
+      axios({
+        method: METHODS.GET,
+        url: `${process.env.REACT_APP_MERAKI_URL}/certificate`,
```
jschanker, line 165: @kirithiv2000 Making the API call directly as follows (removing the token):

```
fetch('https://api.merakilearn.org/certificate', {headers: {"Authorization": 'xxx', "accept": "application/json"}}).then(res => console.log(res));
```
is resulting in an Internal Server Error for me. <s>Took a quick look and this line here: https://github.com/navgurukul/sansaar/blob/dev-copy/lib/services/generateCertificate.js#L86
will cause an error (can't call `toString` on `null`) and isn't used so it can be removed</s>

src/components/PathwayCourse/index.js
```diff
@@ -230,6 +279,22 @@ function PathwayCourse() {
...
         maxWidth="lg"
       >
+        <Modal
+          open={openModal}
+          aria-labelledby="modal-modal-title"
+          aria-describedby="modal-modal-description"
+          onClose={handleModal}
+        >
+        <Box sx={modalStyle}>
+          <Typography sx={{fontSize: "32px", fontWeight: "600"}}>{`${pathwayCourseData?.pathway}  Certifitcate`}</Typography>
```
jschanker: Fixed typo:
```suggestion
          <Typography sx={{fontSize: "32px", fontWeight: "600"}}>{`${pathwayCourseData?.pathway}  Certificate`}</Typography>
```

src/components/PathwayCourse/index.js
```diff
@@ -87,13 +91,32 @@ function PathwayCourse() {
...
   const isActive = useMediaQuery("(max-width:" + breakpoints.values.sm + "px)");
+  const classes = useStyles({isActive});
   const params = useParams();
   const pathwayId = params.pathwayId;
   const [completedPortionJason, setCompletedPortionJason] = useState({});
   // const [loading, setLoading] = useState(true);
   // const [enrolledBatches, setEnrolledBatches] = useState(null);
+  const [openSnackbar, setOpenSnackbar] = useState(false);
+  const [openModal, setOpenModal] = useState(false);
+  const [completedAll, setCompletedAll]= useState(false);
```
jschanker: In general, state should be kept as small as possible. Since this could easily be gotten from `completedPortion`, we should not use `useState` for `completedAll` here. Instead you can just define a variable initialized to a value that uses it.

src/components/PathwayCourse/index.js
```diff
@@ -466,20 +531,40 @@ function PathwayCourse() {
...
             </Grid>
-          </Grid> */}
+            <Snackbar
+              open={openSnackbar}
+              autoHideDuration={6000}
+              message={`Please complete all the courses to unlock ${pathwayCourseData?.pathway}  certificate`}
+              onClose={handleSnackbar}
+              ContentProps={{
+                sx: {
+                  background: "#2E2E2E",
```
jschanker: Reference theme color here

src/components/PathwayCourse/index.js
```diff
@@ -466,20 +531,40 @@ function PathwayCourse() {
...
+              }
               <Typography sx={{mt:2}} variant="body1" mb={2}>
-                {pathwayCourseData?.pathway} Certificate (Locked)
+                {pathwayCourseData?.pathway} Certificate
               </Typography>
             </Grid>
-          </Grid> */}
+            <Snackbar
+              open={openSnackbar}
+              autoHideDuration={6000}
```
jschanker: Should be referencing constants here for consistency in message display durations across the app or may be better to make this its own component to also ensure consistency of styles given in `ContentProps`.

src/components/PathwayCourse/index.js
```diff
@@ -466,20 +531,40 @@ function PathwayCourse() {
...
+              message={`Please complete all the courses to unlock ${pathwayCourseData?.pathway}  certificate`}
```
jschanker: Should use message constants via the `Message` component here to allow support for other languages. (https://github.com/navgurukul/bhanwari-devi/blob/main/src/components/common/Message/index.js)

src/components/PathwayCourse/styles.js
```diff
@@ -104,6 +104,25 @@ const useStyles = makeStyles((theme) => ({
...
+    padding: "8px 16px !important",
+    borderRadius: "8px !important",
+    backgroundColor: "#48A145 !important",
+    color: "#FFFFFF !important",
+    width: "295px !important",
+  },
+  pdfFrame: {
+    height: "400px",
+    width: "100%",
+    margin: "32px 0 32px 0",
```
jschanker:
```suggestion
    margin: "32px 0",
```

src/components/PathwayCourse/styles.js
```diff
@@ -104,6 +104,25 @@ const useStyles = makeStyles((theme) => ({
...
+  certButtons: {
+    display: "flex",
+    justifyContent: "space-around",
+    marginTop: "32px"
+  },
+  greenButton: {
+    padding: "8px 16px !important",
+    borderRadius: "8px !important",
+    backgroundColor: "#48A145 !important",
+    color: "#FFFFFF !important",
```
jschanker, line 116: Reference theme colors here to allow for changes

src/components/PathwayCourse/index.js
```diff
@@ -466,20 +531,40 @@ function PathwayCourse() {
...
+              onClose={handleSnackbar}
+              ContentProps={{
+                sx: {
+                  background: "#2E2E2E",
+                  fontWeight: "400",
+                  fontSize: "18px",
```
jschanker: Using fixed font sizes for fonts is bad for accessibility: What if the user is visually impaired?

src/components/PathwayCourse/index.js
```diff
@@ -87,13 +91,32 @@ function PathwayCourse() {
   const user = useSelector(({ User }) => User);
   const dispatch = useDispatch();
   const { pathwayCourse } = useSelector((state) => state.Pathways);
-  const classes = useStyles();
   const isActive = useMediaQuery("(max-width:" + breakpoints.values.sm + "px)");
+  const classes = useStyles({isActive});
   const params = useParams();
   const pathwayId = params.pathwayId;
   const [completedPortionJason, setCompletedPortionJason] = useState({});
```
jschanker: Recommend you rename `completedPortion` because it's not JSON, it's an object.

jschanker (commented):
Initial comments

src/components/PathwayCourse/index.js
jschanker: Realized that was probably deliberate to force an error to catch since it was in a `try` block. But it crashes for me when a direct request is made.

src/components/PathwayCourse/styles.js
itsomkathe: There are some MUI version issues in the project. So cannot reference the theme prop.

src/components/PathwayCourse/index.js
itsomkathe: I guess translation happens with the help of keys in the `msg` folder. How will it translate a sentence? Haven't really used `Message` before.

src/components/PathwayCourse/index.js
itsomkathe: Have you marked the course as complete through API?

src/components/PathwayCourse/index.js
jschanker: Yes, and you can set the constant to be a complete sentence or phrase. If you need to include a placeholder for a JavaScript expression such as you do here with `pathwayCourseData?.pathway`. just use %1 for the first such placeholder, %2 for the second, etc. in the string literal. Then you can pass the value you want to use for it in the `args` prop. So, for example, here you'd use something like `en.CERTIFICATE_COMPLETION_WARNING = "Please complete all the courses to unlock %1 certificate"` in [en.js](https://github.com/navgurukul/bhanwari-devi/blob/main/src/msg/en.js) and the Hindi translation in `hi.js`. Then could use the following here:
```suggestion
              message={<Message constantKey="CERTIFICATE_COMPLETION_WARNING" args={[pathwayCourseData?.pathway]} />}
```

src/components/PathwayCourse/index.js
jschanker: > Have you marked the course as complete through API?

No, so the request should produce an error response here, I just wasn't sure if it should be a generic 500 Internal Server Error.

src/components/PathwayCourse/index.js
itsomkathe: Thanks for the crisp explanation!

itsomkathe:
The `react-pdf` package that we are using to show PDF as an image is having some issues, which are not solved by the maintainers yet. So as of now, this solution cannot be used.

You can read this thread: https://github.com/wojtekmaj/react-pdf/issues/1043

 Below is the screenshot attached

Capture

PR #787, opened 2022-11-05
Fixed: Checkbox, Pathway Filters, default Pathway selection

amansharmma:
Fixed: Checkbox, Pathway Filters, default Pathway selection.

Contained button for Selected Pathway
If selected multiple rows are the header row becomes '-' checkbox when clicking on '-' checkbox, all selected rows should be deselected.
No combined data of Python and English (By default, Python selected)

Screenshot from 2022-11-05 14-58-57


src/msg/en.js
```diff
@@ -11,7 +11,7 @@ en.VOLUNTEER = "Volunteer";
 en.ADMIN = "Admin";
 en.PARTNER = "Partner";
 en.STUDENTS = "Students";
-en.VOLUNTEERS = "Volunteers";
+en.VOLUNTEERS = "Tutors";
```
jschanker, line 14: Remember you'd need to update the Hindi translation in `hi.VOLUNTEERS` as well. But it might be better to instead create a new constant for `TUTORS` in both `en.js` and `hi.js` and then do an (automated) search of all files to look for all references to `VOLUNTEERS`, changing the ones that should be changed to `TUTORS`. This is because we still might want to reference the constant for Volunteers in another place.

src/components/TutorDashboard/index.js
```diff
@@ -150,6 +150,8 @@ function Tutor(props) {
   const [endTime, setendTime] = useState("");
   const [slicedStudents, setSlicedStudents] = useState([]);
   const [statusValue, setStatusValue] = useState("");
+  const [pathwayClicked, setPathwayClicked] = useState(true);
+  const [pathwayTwoClicked, setPathwayTwoClicked] = useState(false);
```
jschanker: You want to avoid having dependent state like this where one is always the opposite of the other because (a) we have to remember to update both together (or create a function that does this and remember to call that function) and (b) given we're currently using React < 18, this could unnecessarily cause extra re-renders (See: https://dev.to/this-is-learning/automatic-batching-in-react-18-273h#:~:text=Batching%20is%20when%20React%20groups,batched%20in%20React%20by%20default.) You could do the following but there is no reason to have `pathwayTwoClicked` anyway.

```suggestion
  const [pathwayClicked, setPathwayClicked] = useState(true);
  const pathwayTwoClicked = !setPathwayClicked;
```

src/components/TutorDashboard/index.js
```diff
@@ -405,24 +407,29 @@ function Tutor(props) {
...
           <Grid item>
             <Button
-              variant="contained"
+              // "contained"
               className={classes.python}
               onClick={() => {
                 setVolunteer(filterPathway(1, cacheVolunteer));
                 setSelectedPathway("Python");
+                setPathwayClicked(true);
+                setPathwayTwoClicked(false);
```
jschanker: Now after making the above suggestion, you only need to set pathway clicked, the other one will automatically update.
```suggestion
```

src/components/TutorDashboard/index.js
```diff
@@ -405,24 +407,29 @@ function Tutor(props) {
...
           </Grid>
           <Grid item>
             <Button
-              variant="outlined"
               className={classes.learningTrack2}
               onClick={() => {
                 setVolunteer(filterPathway(2, cacheVolunteer));
                 setSelectedPathway("Spoken English");
+                setPathwayClicked(false);
+                setPathwayTwoClicked(true);
```
jschanker: Here as well although the code for the Button is duplicated and should be factored out.
```suggestion
```

src/components/TutorDashboard/index.js
```diff
@@ -405,24 +407,29 @@ function Tutor(props) {
...
               }}
+              variant={pathwayTwoClicked ? "contained" : "outlined"}
```
jschanker: Instead of how you wrote this, you can remove `pathwayTwoClicked` and just write the following:
```suggestion
              variant={pathwayClicked ? "outlined" : "contained"}
```

PR #804, opened 2022-11-15
Review of feat/Content Editor

jschanker:
Review of https://github.com/navgurukul/bhanwari-devi/tree/feat/contentEditor

@Poonam-Singh-Bagh


src/components/PathwayExercise/ContentEdit/index.js
```diff
@@ -0,0 +1,450 @@
...
+              component: "questionCode",
+              type: "python",
+              title: "",
+              value: "",
+            });
+          }
+        }
+        setId(res.data.course.exercises[exerciseId].id);
+        setCourse(res.data.course.exercises[exerciseId].content);
+        const Date = res.data.course.exercises[exerciseId].updated_at;
```
jschanker, line 182: Avoid using the variable name `Date` as this will prevent any future use of the built-in `Date` functionality in this scope and can cause confusion (e.g., `Date.now()` wouldn't work) . Also, the convention would be to use all lowercase letters to represent a `Date` string.

src/components/PathwayExercise/ContentEdit/index.js
```diff
@@ -0,0 +1,450 @@
...
+        const date = Date.split("T")[0].replace(
+          /(\d{4})-(\d{1,2})-(\d{1,2})/,
+          function (match, y, m, d) {
+            return d + "/" + m + "/" + y;
+          }
+        );
+        var longDateStr = moment(date, "D/ M/Y").format("dddd D MMMM Y");
```
jschanker, line 189: Try to avoid using lower-level date manipulation like this and the use of the deprecated `moment`.  Additionally, remember that we want to make calls to functions from https://github.com/navgurukul/bhanwari-devi/blob/Feat/contentEditor/src/common/date.js so we use centralized date functionality. This makes us less likely to duplicate date-related code. It also allows us to easily migrate from `date-fns` to some other date-related library in the future if desired, without needing to touch component code that depends on it. Finally, it allows us to make changes in one place, if say, the format dates are stored were changed on the back-end as has been done in the past.

src/components/PathwayExercise/ContentEdit/index.js
```diff
@@ -0,0 +1,450 @@
...
+                        placeholder="Question"
+                        color="primary"
+                        className={classes.textarea}
+                        value={course[index].value}
+                        onChange={(e) => {
+                          var temp = [...course];
+                          temp[index].value = e.target.value;
+                          setCourse(temp);
+                        }}
+                      />
```
jschanker, line 268: Duplicated below, factor out into a component? However, for future reference, rather than writing the code first and then trying to factor out common code into components, it'll be good to start with a mock and then break it into a hierarchy as outlined here: https://reactjs.org/docs/thinking-in-react.html#start-with-a-mock. The point is that components should be small and have only one reason to change to maximize reusability and make them easier to test and debug.

src/components/PathwayExercise/ContentEdit/editor.js
```diff
@@ -0,0 +1,276 @@
...
+        height: 320,
+        width: 580,
+        // withHeadings: true,
+        // content: content,
+      },
+    };
+  });
+
+  let blocks = courseData.filter((item, index) => {
+    const stringifiedItem = JSON.stringify(item);
```
jschanker, line 113: Using `JSON.stringify` repeatedly for large objects can negatively affect performance and is unnecessary in this case since we only need to compare the `items` properties of the objects. Here, we can use `===` for that as `items` will refer to the same objects. Additionally, note that while not relevant here, there could be unexpected results with this approach so when comparing objects for objects, it's better to use built-in tested functions such as Lodash's `isEqual` (https://lodash.com/docs/4.17.15#isEqual):
```
JSON.stringify({f: x => 2*x}) === JSON.stringify({}) // true, function isn't serialized
JSON.stringify({a: 42, b: 0}) === JSON.stringify({b: 0, a: 42}) // false, order of keys is different
```

)

jschanker (commented):
Some initial comments

src/components/PathwayExercise/ContentEdit/editor.js
Poonam-Singh-Bagh: I had tried with `isEqual` Lodash but got confused that, which 2 objects I needed to compare.

jschanker: It should be the same ones for which you're comparing the values of `JSON.stringify`, so below you could use `_.isEqual(obj, item)` instead.

src/components/PathwayExercise/ContentEdit/index.js
Poonam-Singh-Bagh: Can we change `16/11/2022` to `Wednesday 16 November 2022` (`"dd/MM/yyyy"` to `"dddd D MMMM Y"`) using this [function](https://github.com/navgurukul/bhanwari-devi/blob/Feat/contentEditor/src/common/date.js#L235-L237) 
I replaced the code `Date.split` with this function by calling it, but  unable to do the same for `moment(date, "D/ M/Y").format("dddd D MMMM Y");`

jschanker: Sure, but you would call it on `updated_at`, not `date`. So, it'd be `formatInUtc(updated_at, "EEEE dd MMMM yyyy")` (See: https://date-fns.org/v2.0.0-alpha.25/docs/format for the list of format tokens for `date-fns`, which can differ from those from `moment`). However, is there a reason to format it in UTC instead of the user's local time? For example, if it were last modified at 1 AM UTC on 11/17, would we want it to read that it was last modified on 11/16 for me since I'm currently 5 hours behind (11/16 at 8pm EST)? If so, you can use `format` instead.

Poonam-Singh-Bagh: There isn't a reason in the course editor but in create class form as backend wants it.

Poonam-Singh-Bagh: `formatInUtc(updated_at, "EEEE dd MMMM yyyy")` here `EEEE` stands for day?

jschanker:
@Poonam-Singh-Bagh Missed these comments before (https://github.com/navgurukul/bhanwari-devi/pull/804#discussion_r1024935809 and https://github.com/navgurukul/bhanwari-devi/pull/804#discussion_r1024944101):

> There isn't a reason in the course editor but in create class form as backend wants it.

But aren't we just formatting the date for the user here? In this case, I think we should be using the user's local time or at least mention that it's in UTC. Otherwise, it could be confusing to the user. As for the back-end, we should be sending a timestamp `updated_at`, not an English date to the back-end. But even if we did need to send it this way, this should not affect how we display it to users, which should be independent of its internal representation.

> formatInUtc(updated_at, "EEEE dd MMMM yyyy") here EEEE stands for day?

Yes, according to the documentation `EEEE` is `Day of week (formatting)`. `eeee` will be `Local day of week (formatting)`.

jschanker:
@Poonam-Singh-Bagh To follow up, I just checked and it doesn't seem that we're sending the last edited date to the back-end: https://github.com/navgurukul/bhanwari-devi/blob/f13aa1819f24277257e5a136f7f7bda2e09d2154/src/components/PathwayExercise/ContentEdit/index.js#L126-L128
Is that correct? Even though the back-end sends the timestamp in UTC, the `format` function will show it correctly in the user's local time. For example, try entering `new Date('2022-11-20T09:05:05.302Z')` in the console. This displays as "Sun Nov 20 2022 04:05:05 GMT-0500 (Eastern Standard Time)" for me, but if you tried it, it should display as something like "Sun Nov 20 2022 14:35:05 GMT+0530 (Indian Standard Time)." So, you should be able to use `format` instead of `formatInUtc`.

jschanker:
Closing because review against https://github.com/navgurukul/bhanwari-devi/tree/feat/contentEditor will be here: https://github.com/navgurukul/bhanwari-devi/pull/810/

PR #810, opened 2022-11-20
Review of feat/Content Editor

jschanker:
Review of https://github.com/navgurukul/bhanwari-devi/tree/feat/contentEditor, continuing from https://github.com/navgurukul/bhanwari-devi/pull/804. Making a separate PR here because the PR was against dev instead of `feat/contentEditor` where the changes are being made.

1. Upgraded packages
3. ...

@Poonam-Singh-Bagh


src/components/PathwayExercise/ContentEdit/index.js
```diff
@@ -0,0 +1,487 @@
...
+              value: "",
+            });
+          }
+        }
+        setId(res.data.course.exercises[exerciseId].id);
+        setCourse(res.data.course.exercises[exerciseId].content);
+        const updated_at = res.data.course.exercises[exerciseId].updated_at;
+        const longDateStr = formatInUtc(updated_at, "EEEE dd MMMM yyyy");
+        // console.log("longDateStr", longDateStr);
+        setUpdatedOn(longDateStr);
```
jschanker: As noted in https://github.com/navgurukul/bhanwari-devi/pull/804#issuecomment-1321073712 and https://github.com/navgurukul/bhanwari-devi/pull/804#issuecomment-1321080007, you should be able to use `format` here instead. Also, we don't need the extra variable declaration since it's used immediately and not used anywhere else.
```suggestion
        setUpdatedOn(format(updated_at, "EEEE dd MMMM yyyy"));
```

src/components/PathwayExercise/ContentEdit/styles.scss
```diff
@@ -0,0 +1,31 @@
+.embed-tool__caption {
+  display: none;
+}
+
```
jschanker, line 4: When you need to use a hack to make something work, it might be useful to document it in the issues and then reference it in a comment. This way we can keep track of them. Doing so will also allow a potential contributor to search the issues and perhaps make a PR if they think of a better solution that comes up in the future.
 
```suggestion

// Issue # 811: Hack to remove image caption
```

src/components/PathwayExercise/ContentEdit/editor.js
```diff
@@ -0,0 +1,282 @@
...
+        stretched: true,
+        embed: item.component === "youtube" && youtube,
+        source: item.component === "youtube" && youtube,
+        service: item.component === "youtube" && item.value && item.component,
+        height: 320,
+        width: 580,
+        // withHeadings: true,
+        content: content,
+      },
+    };
```
jschanker, line 114: The way this is set up, our blocks will be filled with unused properties. For example, the `embed` and `source` properties won't be used for paragraphs yet they still appear in the blocks. Additionally, the logic for the conversions into different types of blocks gets mixed together and has multiple reasons to change. We could make this cleaner, by having an object whose keys are component types and values are functions that return the corresponding block object. For example, the object could have a property with key `youtube` and a function as follows (See https://stackoverflow.com/questions/830596/what-type-of-id-does-youtube-use-for-their-videos)
```
toEditorBlock: (item) => {
  const youtube = "https://www.youtube.com/embed/" + 
    (/=([\w\-]+)/.exec(item.value)?.[1] || item.value);
  return {
    type: "embed",
    data: {
      embed: youtube,
      source: youtube,
      service: item.value && item.component,
      height: 320,
      width: 580,
  };
}
```

jschanker (commented):
Few more comments for now

PR #829, opened 2022-12-05
commented Github pathwaydata , reduced unnecery api calls

kartiks26:
1. Resolved Bug of Github URL failing in giving response causing late response . Commented code snippet and tested flow elsewhere to check if nothing is breaking
2.  Removed Unnecessary calling of Batch data and class data api like on javaScript and other
3. Removed Unnecessary calling of Learning track api like in open course , residential pathways.


jschanker:
@kartiks26 Not sure why the request to the raw data file is failing? It does work for me. But its contents haven't been updated for a while because the GitHub action to do so has been turned off,

kartiks26:
@jschanker its failing for me everytime not sure why

jschanker:
> @jschanker its failing for me everytime not sure why

@kartiks26 Does it fail for you on https://bhanwari-devi.vercel.app/ or https://merakilearn.org/ ? If it's failing locally, is it a timeout issue? Assuming it's not CORS if it's causing a noticeable delay as the error should come immediately.

kartiks26:
@jschanker 
image
Failing for main too

jschanker:
@kartiks26 This is strange, I tried it on my home computer and at school, and it worked in both places. I can't see the reason why it failed from the image though. Does the console error or clicking on one of the failures to see the request/response show anything? If you're passing any extra request header, that would make it fail, but as you're loading the same code from merakilearn.org that I am (and it shouldn't be dependent on your login), I'm at a loss. Maybe a browser extension you're using?

PR #850, opened 2022-12-14
Content Drawer: Preserve Scroll Position

itsomkathe:
Content List, when changed content, was scrolling back to the top. 

1. Save the scroll position in `localStorage` every time it is changed.
2. When the list gets re-rendered, scroll to the saved position.
3. Remove the position from `localStorage` when exited.
4. Used debouncing to improve performance.


saquibaijaz:
@komalahire @itsomkathe 
Check out the attached screencast.
When we click on any item in the content list, it comes on two lines as it gets bold, and an arrow symbol is added at the start. 
Maybe increase the content list space, so it remains in a single line. Or maybe do not make it bold and add an arrow, just change the item's background colour.

In whichever way you do it, make sure it's in a single line, even after clicking.

https://user-images.githubusercontent.com/111858763/208891760-cfe02bc5-5363-4321-8d72-dc53f7eb44eb.mp4

PR #924, opened 2023-03-04
Feat/code editor rtc

RTC-13:
Coming soon


src/components/CodeEditor/CodeMirror.js
```diff
@@ -0,0 +1,42 @@
...
+        }}
+        onBeforeChange={handleEditorChange}
+        value={value}
+        className="code-mirror-wrapper"
+        options={{
+          lineWrapping: true,
+          lint: true,
+          mode: "python",
+          lineNumbers: true,
+          indentUnit: 4,
```
jschanker: We'll want `indentUnit` (https://codemirror.net/5/doc/manual.html#option_indentUnit) to match what we replace `&emsp;` with, which is currently one space. Otherwise, there will be inconsistent indentation when adding new lines in a code block such as a while loop. Additionally, code will not highlight correctly. For example, for the below code block, print appears in red (instead of blue as it would if it were indented four spaces). However, if we click enter after the line `i = i + 1`, the code will be indented 4 spaces, which will cause errors due to the inconsistency.
```
i = 0
while i < 100:
 print("I will be printed again and again:)")
 i = i+1
```
To maintain this consistency in the case of changes, I would add a constant like [INDENT](https://github.com/google/blockly/blob/42fde0f81b5d89c989741c6db41726bce7d2d781/core/generator.ts#L61-L65) to the main constants file (https://github.com/navgurukul/bhanwari-devi/blob/feat/code-editor-rtc/src/constant.js), set it to the current value of 4 (https://github.com/navgurukul/sansaar/blob/main/lib/courseParser/index.js#L1135), and then also reference that same constant in the `ExerciseContent` file.

```suggestion
          indentUnit: INDENT,
```

src/components/CodeEditor/PythonEditor.js
```diff
@@ -0,0 +1,154 @@
...
+    <Box
+      className="PythonCodeEditor"
+      sx={{
+        my: 3,
+      }}
+    >
+      <Box
+        sx={{
+          p: "2px",
+          border: "1px solid #6D6D6D",
```
jschanker: Reference theme colors instead of hardcoding `#6D6D6D` in each place because these can change:
https://github.com/navgurukul/bhanwari-devi/blob/f0b79712ebe46fc4ade63f68fde662b36319b70a/src/theme/theme.js#L71

src/components/CodeEditor/PythonEditor.js
```diff
@@ -0,0 +1,154 @@
...
+  initialCodeEditorValue,
+  disableEditing,
+  disableRun,
+}) => {
+  let { runPython, stdout, stderr, isLoading, isRunning } = usePython();
+  const [pythonEditorCode, setPythonEditorCode] = useState(
+    initialCodeEditorValue
+  );
+  const [codeRan, setCodeRan] = useState();
+  let [codeExecuted, setCodeExecuted] = useState(stdout, stderr);
```
jschanker: Use `const` here as neither variable is changing during a single render. Also, `useState` only takes one input, the initial value. By using `stdout, stderr`, you're using the [comma operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comma_Operator) which will return the last of these values, mainly `stderr`.

jschanker: Why are these state variables necessary? Why can't we just use `isLoading` and `isRunning` since every time the values of these variables change, the component should be re-rendered.

src/components/CodeEditor/PythonEditor.js
```diff
@@ -0,0 +1,154 @@
...
+              >
+                Reset
+              </Button>
+              <Button
+                endIcon={<ArrowRightIcon />}
+                disabled={
+                  isLoading ||
+                  isRunning ||
+                  (codeRan === pythonEditorCode && codeExecuted)
+                }
```
jschanker: Why does `(codeRan === pythonEditorCode && codeExecuted)` have any determination in whether the button is disabled here? It seems to me that you'd only want to disable it when it's loading or the code is running.
```suggestion
                disabled={isLoading || isRunning}
```

src/components/CodeEditor/PythonEditor.js
```diff
@@ -0,0 +1,154 @@
...
+
+import CodeMirrorEditor from "./CodeMirror";
+
+import ArrowRightIcon from "@mui/icons-material/ArrowRight";
+import RestartAltIcon from "@mui/icons-material/RestartAlt";
+import CableIcon from "@mui/icons-material/Cable";
+
+import { Box, Button, Grid, Typography } from "@mui/material";
+
+const PythonEditor = ({
```
jschanker, line 22: Document this component using JSDoc (https://jsdoc.app/)

src/components/CodeEditor/PythonEditor.js
```diff
@@ -0,0 +1,154 @@
...
+            >
+              <Button
+                startIcon={<RestartAltIcon />}
+                disabled={
+                  isLoading ||
+                  (initialCodeEditorValue === pythonEditorCode && !codeExecuted)
+                }
+                variant="outlined"
+                onClick={() => {
+                  console.log(initialCodeEditorValue);
```
jschanker: Remove `console.log` statements before merging. 
```suggestion
```

src/components/PathwayExercise/ExerciseContent/index.js
```diff
@@ -248,26 +232,31 @@ const RenderContent = ({ data, exercise }) => {
...
-          </Grid>
         </Box>
       </div>
     );
   }
+
+  if (data.component === "code" && data.type === "python") {
+    const pythonCode = data.value
+      .replace(/<br>/g, "\n")
+      .replace(/&emsp;/g, " ")
```
jschanker: Consider adding an object to the constants file that has all of these replacements (See a related comment of why here: https://github.com/navgurukul/sansaar/pull/469#discussion_r771807014). Here, `&emsp;` would be replaced with a number of spaces equal to the value of the `INDENT` constant mentioned earlier.

jschanker (commented):
Some initial comments

src/components/CodeEditor/CodeMirror.js
RTC-13: Done.

src/components/CodeEditor/PythonEditor.js
RTC-13: Done. This color is currently under theme.palette.text.secondary, would this make the color volatile. Meaning that it could be changed accidentally thinking it only affects the style of the text?

src/components/CodeEditor/PythonEditor.js
RTC-13: One of the two will be true (have a value), stdout or stderr. Is this a bad approach?

src/components/CodeEditor/PythonEditor.js
jschanker: @RTC-13 Well, the question is whether we want to have it always match the color of the text, which I'm assuming we do. We should ask Narendra about this to confirm.

jschanker (dismissed):
Future TODO:
1. Currently every code editor will cause a new request to load `Pyodide`. This is inefficient, but should not cause any issues. For the future, we could fix this by moving the `usePython` hook up to an ancestor and passing down the relevant properties so all code editors can run their code without triggering their own requests to load Pyodide.
2. The workaround we used to get `react-py` to work is causing a number of tests to fail: https://github.com/navgurukul/bhanwari-devi/pull/924#issuecomment-1493604832 . We might be able to reduce the number of these errors if we had copied the code as is instead of running it through Babel to transpile to ES2015. This should be fixed once we update packages.

PR #1002, opened 2023-06-13
Integration with strapi

komalahire:
Merd Integration

 Worked on Header, Footer, Student dashboard, and home page, removed hard-coded pathway data from all components, and Integrate with API

@jschanker


src/components/PathwayCourse/index.js
```diff
@@ -132,12 +131,16 @@ function PathwayCourse() {
   const [openSnackbar, setOpenSnackbar] = useState(false);
   const [openModal, setOpenModal] = useState(false);
   const [certificate, setCertificate] = useState("");
-  let completedAll = pathwayId==8?completedPortion?.total>=80: completedPortion?.total === 100;
+  let completedAll =
+    pathwayId == 8
+      ? completedPortion?.total >= 80
+      : completedPortion?.total === 100;
```
jschanker: Few comments here:
1. Using `==` will add lint. Maybe use `parseInt(pathwayId) === 8` instead.
2. We don't want to be hardcoding ids or minimum completed percentages here as they can change. We could remove the reference to the id here of `8` and the minimum completion percentage of `80` if we instead have an object mapping ids to the values necessary to complete them. If the id isn't present, we could use `100` (e.g., `const MINIMUM_COMPLETE_PERCENTAGES = {'8': 80}`) . This would allow us to easily change minimum completed portions of other pathways as well without touching the code. However, it would be better to get this map of completed percentages from the back-end. For now (after importing this constant):
```suggestion
    completedPortion?.total >= (MINIMUM_COMPLETE_PERCENTAGES[pathwayId] || 100);
```

src/components/PathwayCourse/index.js
```diff
@@ -334,8 +339,7 @@ function PathwayCourse() {
                 align="center"
                 className={classes.classTitle}
                 variant="subtitle1"
-                bgcolor="#E9F5E9"
-              >
+                bgcolor="#E9F5E9">
```
jschanker: Reference theme color here?

src/components/PathwayCourse/index.js
```diff
@@ -132,12 +131,16 @@ function PathwayCourse() {
   const [openSnackbar, setOpenSnackbar] = useState(false);
   const [openModal, setOpenModal] = useState(false);
   const [certificate, setCertificate] = useState("");
-  let completedAll = pathwayId==8?completedPortion?.total>=80: completedPortion?.total === 100;
+  let completedAll =
```
jschanker: `completedAll` is a little misleading as a variable name if you only need to complete a minimum percentage. Also, use `const` here since the variable is not reassigned.

jschanker (commented):
Some initial comments. Will continue reviewing later.

jschanker:
Issue with the dropdown image sizes?
image



What is your screen size @jschanker

src/components/NewUserDashbord/index.js
```diff
@@ -44,22 +44,13 @@ const NewUserDashbord = () => {
...
-        pathwayData.push(obj);
-      }
-    });
+  const MiscellaneousCourses = data?.pathways.filter(
+    (item) => item.name === "Miscellaneous Courses"
+  );
+
+  const pathwayData = data?.pathways
+    .filter((item) => item.name !== "Miscellaneous Courses")
+    .concat(MiscellaneousCourses);
```
jschanker: Is the reason for this code to include the Miscellaneous courses pathway at the end? We'll probably want to avoid explicit mention of the name in case we decide to change its name or offer translations into other languages. Perhaps we could `import` and use `PATHWAY_INFO` to determine whether it should be at the end of not. Also, we should probably rename the variable to indicate that it's a pathway name and not the collection of miscellaneous courses such as Web Scraping 101.

For example:
```suggestion
  const miscellaneousPathway = data?.pathways.filter(
    (pathway) => PATHWAY_INFO.some((miscPathway) => pathway.name === miscPathway.name)
  );

  const pathwayData = data?.pathways
    .filter((pathway) => !miscellaneousPathway.includes(pathway))
    .concat(miscellaneousPathway);
```

src/pages/Home/index.js
```diff
@@ -161,23 +161,9 @@ function Home() {
...
-          image: pathway.image || pathway.logo,
-          link: pathway.path || null,
-          type: "internal",
-        };
-        pathwayData.push(obj);
-      }
-    });
+  const MiscellaneousCourses = data?.pathways.filter(item => item.name === 'Miscellaneous Courses')
 
+  const pathwayData = data?.pathways.filter(item => item.name !== 'Miscellaneous Courses').concat(MiscellaneousCourses)
```
jschanker: Similar comment here about explicit mention of "Miscellaneous courses."

src/components/PathwayCourse/index.js
```diff
@@ -132,12 +131,16 @@ function PathwayCourse() {
...
+    pathwayId == 8
+      ? completedPortion?.total >= 80
+      : completedPortion?.total === 100;
   // let completedAll = true
   const [loader, setLoader] = useState(false);
-  const displayCert = pathwayId == 1 
+  const displayCert = pathwayId == 1;
```
jschanker: Similar comment to https://github.com/navgurukul/bhanwari-devi/pull/1002/files#r1229700374. There's the lint issue of using `==` and also we want to avoid explicit references to pathway ids in case they change. So, let's maybe add a line `const IDS_OF_PATHWAYS_WITH_CERTIFICATES = ['1'];` to the constants file and then we could do the following. This would also have the added benefit of allowing us to add pathways with certificates in the future by merely updating the constants file, without touching the code. Again, it'd be good to get this information from the back-end at some point.
```suggestion
  const displayCert = IDS_OF_PATHWAYS_WITH_CERTIFICATES.includes(pathwayId);
```

src/components/PathwayCourse/index.js
```diff
@@ -395,29 +397,28 @@ function PathwayCourse() {
...
-                      >
+                        mb={isActive ? 16 : 30}>
                         {pathwayCourseData.code !== "SHCEL" && (
                           <Typography
                             variant="body2"
                             className={classes.cardSubtitle}
-                            sx={{ textAlign: isActive && "center", pb: "8px" }}
-                          >
+                            sx={{
+                              textAlign: isActive && "center",
```
jschanker: Not really part of this PR, but it would be good to remove `isActive` as possible to have the styles update with CSS media queries instead of re-renders.
```suggestion
                              textAlign: {xs: "center", sm: "left"},
```

src/pages/Home/PathwayCard.js
```diff
@@ -4,45 +4,40 @@ import { Link } from "react-router-dom";
...
+          {logo && (
             <CardMedia
               component="img"
               src={
-                image.includes('https')
-                  ? image
-                  : require("./assets/" + image + ".svg")
+                logo.includes("https")
+                  ? logo
+                  : require("./assets/" + logo + ".svg")
```
jschanker, line 29: May want to have a function that returns the paths of the logos to use based on what comes from the back-end. This would allow for changes (e.g., if logos aren't going to be in assets folder) without editing the logic in multiple components.

Poonam-Singh-Bagh:
> Issue with the dropdown image sizes? image

src/components/PathwayCourse/index.js
Poonam-Singh-Bagh: For now, we are going for `pathway.code`. After it's released it will come from the backend

src/components/NewUserDashbord/index.js
komalahire: Okay, I will change it but I will ask the backend to send this course from the backend only.

jschanker:
> Issue with the dropdown image sizes? image
> 
> What is your screen size @jschanker

@Poonam-Singh-Bagh Screen or browser window width? It happens at any browser window width for me. This seems to be due to some of the images being larger than the ones that they're replacing. For example, the Typing one located at https://merakilearn.s3.ap-south-1.amazonaws.com/Meraki+track+icons/typing/type%3Dtyping_xxxhdpi.png is 128 x 128 as opposed to the Python one at https://merakilearn.s3.ap-south-1.amazonaws.com/Meraki+track+icons/python/Python.png, which is only 48 x 48.

Issue #803, opened 2022-11-14
Inconsistent menus in mobile sidebar


jschanker:
**Describe the bug**
The Change Roles/Public menu dropdowns aren't consistent in the sidebar. The public ones gradually show/hide their dropdowns (accordion) on clicks but not on hovers and stay in the flow of the sidebar. The change roles view one appears in its own dropdown modal and also responds to hovers. This is because the public menus are handled with different components when the sidebar is open (`MobileDropdown`) whereas the `ChangeRolesView` menu uses the `Dropdown` regardless of whether it appears at the top or in the sidebar.

**To Reproduce**
1. When logged in, switch to the student role.
2. Make the browser window narrow enough to see the hamburger menu and click it.
3. Expand the student dropdown and then click/hover over the Switch Views menu button.

**Expected behavior**
Consistency in user menus, both accordion (preferable @saquibaijaz ?) or both modals.

**Screenshots**
![image](https://user-images.githubusercontent.com/9159008/201709378-14875f2d-b74a-42b8-a2ad-18a367a7df36.png)

**Additional context**
Will be resolved in PR #740 once UI decision is made.

PR #1018, opened 2023-07-07
Header refactor

Ravi140891:
Refactored the header components and optimized them to make them more memory efficient and make code readable.


src/components/Header/ChangeRolesView/index.js
```diff
@@ -39,109 +15,93 @@ function ChangeRole({ isToggle, role, setRoleView, roleView, uid }) {
...
+  const handleRoleChange = (selectedRole) => {
+    setRoleView(selectedRole);
+  };
+
+  if (roles.length > 1) {
+    const menuContents = roles.map((role) => (
+      <ChangeRole
+        role={role}
+        key={role.key}
+        isToggle={roles.length > 2}
```
jschanker, line 78: Toggling (no dropdown) should only occur when the number of roles is exactly equal to 2. It should look something like this (while hovering over):

image

```suggestion
        isToggle={roles.length === 2}
```

src/components/Header/ChangeRolesView/index.js
```diff
@@ -39,109 +15,93 @@ function ChangeRole({ isToggle, role, setRoleView, roleView, uid }) {
...
-    localStorage.getItem(ID_TO_SELECTED_ROLE_MAP_KEY)
-  )?.[uid]; // || localStorage.getItem(SELECTED_ROLE_KEY);
-  const hasLastSelectedRole = roles.some(
-    (role) => role.key === lastSelectedRoleKey
-  );
-  const [roleView, setRoleView] = React.useState(
-    hasLastSelectedRole ? lastSelectedRoleKey : defaultRole?.key
-  );
-  const otherRole =
-    roles[(roles.findIndex((role) => role.key === roleView) + 1) % 2];
```
jschanker, line 87: This is used for the toggling

src/components/Header/ChangeRolesView/index.js
```diff
@@ -39,109 +15,93 @@ function ChangeRole({ isToggle, role, setRoleView, roleView, uid }) {
...
+        role={role}
+        key={role.key}
+        isToggle={roles.length > 2}
+        setRoleView={handleRoleChange}
+        roleView={roleView}
+        uid={uid}
+      />
+    ));
+
+    if (roles.length > 2) {
```
jschanker, line 85: We also need to check that `leftDrawer` is `true` (meaning the mobile sidebar is open) or we'll get something like this:
image
```suggestion
    if (roles.length > 2 && leftDrawer) {
```

src/components/Header/RoleSpecificHeader/index.js
```diff
@@ -15,9 +11,7 @@ function RoleSpecificHeader({ role, isUniqueRole, leftDrawer, toggleDrawer }) {
 
   const roleSpecificComponentMap = {
     [STUDENT]: <StudentHeader isUniqueRole={isUniqueRole} {...drawerProps} />,
-    [ADMIN]: <AdminHeader {...drawerProps} />,
```
jschanker, line 18: I'm assuming we're getting rid of the admin and partner roles for the Meraki app? In that case, you'll need to remove it from them from `ROLES` in the constant file. Otherwise, if the user has one of these as their first role (or selected role in `localStorage`), it will load a blank header like below. Check that this wouldn't cause other issues.
image

src/components/Header/ChangeRolesView/index.js
```diff
@@ -39,109 +15,93 @@ function ChangeRole({ isToggle, role, setRoleView, roleView, uid }) {
...
+      menu = (
+        <AccordionDropDownMenu textMsgKey="SWITCH_VIEWS">
+          {menuContents}
+        </AccordionDropDownMenu>
+      );
+    } else {
```
jschanker, line 91: Similarly, we need `leftDrawer` to be `false` here (or we don't get the accordion when the sidebar is opened):
image

See this issue here: https://github.com/navgurukul/bhanwari-devi/issues/803
```suggestion
    else if (roles.length > 2) {
```

jschanker (commented):
I started reviewing this PR, and I'm finding some regressions, which I mention in the comments. Also, there's this one where the student mobile sidebar is empty:
image
Given the complexity of the header code with all of the patches we've added to address subtle issues we've had in the past, regressions can be expected during refactoring. Accordingly, I think we need a comprehensive collection of tests.

PR #1035, opened 2023-08-03
merge class

saloni-080601:
merge class for ACB
New design of class.


src/components/Class/ClassCard/index.js
```diff
@@ -261,7 +264,11 @@ function ClassCard({ item, editClass }) {
     <>
       <Card
         elevation={2}
-        sx={{ p: 4, mt: isActive ? 4 : 5 }}
+        sx={{
+          p: 4,
+          mt: isActive ? 4 : 5,
+          background: timeZone ? "#FFF5CC" : "#FAFAFA",
```
jschanker: Reference theme colors here, so if they need to change in the future, we can make the change in one place and be sure that they're the same everywhere across the site.

Also, use `canJoin` instead of `timeZone` to better indicate the variable's meaning.

src/components/Class/ClassCard/index.js
```diff
@@ -310,13 +330,20 @@ function ClassCard({ item, editClass }) {
...
               >
                 <Typography textAlign="center">Edit</Typography>
               </MenuItem>
+              {item.pathway_id === 7 && (
+                <MergeClass
+                  itemID={item.id}
+                  PathwayID={item.pathway_id}
+                  pathwayFilter={pathwayFilter}
+                />
+              )}
```
jschanker: Is the check for the `pathway_id` in this code temporary? If not, let's add a constant in the constants file with an Array of all `pathway_id`'s for which the `MergeClass` component should be shown. For now, it'd just be `[7]`. Then we can instead do a check like `MERGE_CLASSES.includes(item.pathway_id)` instead of `item.pathway_id === 7`. This allows us to add/remove pathways for which `MergeClass` will be shown by modifying a constant fils without touching this code.

src/components/Class/ClassCard/index.js
```diff
@@ -333,11 +360,24 @@ function ClassCard({ item, editClass }) {
...
+            <img
+              className={classes.icons}
+              src={require("../assets/mergeClass.png")}
+              height="26px"
+              width="26px"
+            />
+
+            {item?.merge_class}
+          </Typography>
+        )}
```
jschanker: Similar comment here about referring to a constant Array instead of hardcoding the id in the code.

src/pages/CreateClass/index.js
```diff
@@ -153,115 +162,192 @@ function ToggleClassFormModal() {
...
+
+  useEffect(() => {
+    axios({
+      method: METHODS.GET,
+      url: `${process.env.REACT_APP_MERAKI_URL}/pathways/names`,
+      headers: {
+        accept: "application/json",
+        Authorization: user.data.token,
+      },
+    }).then((res) => {
```
jschanker, line 170: Probably not a big deal because the response data is small but can we not get this information from the response from `pathway/dropdown` which we had to make anyway?

src/components/Class/ClassJoinTimerButton/index.js
```diff
@@ -2,12 +2,14 @@ import React, { useState, useEffect } from "react";
...
 import { Button } from "@mui/material";
 import ExternalLink from "../../common/ExternalLink";
+import ArrowRightAltIcon from "@mui/icons-material/ArrowRightAlt";
 
 function ClassJoinTimerButton({
   startTime = new Date(),
   link,
   joinOnClick,
   buttonType,
+  setTimeZone,
```
jschanker: We can make this property more general to be any function we want to run when the user is able to join. Also, it'd be good to rename it something like `onCanJoin` to better reflect its meaning.

src/components/Class/ClassJoinTimerButton/index.js
```diff
@@ -40,26 +42,37 @@ function ClassJoinTimerButton({
     return () => clearInterval(timer); // cleans up on unmount
   }, [startTime]);
 
+  useEffect((e) => {
+    if (timeRemainingMsg === CAN_JOIN_MSG) {
+      setTimeZone(true);
+    }
+  });
```
jschanker: Without dependencies, this function will run on every render. Not really a problem but it may be better to just move `onCanJoin()` up to line 40 where we're already checking if the user can join. Then we don't need an extra `useEffect`.

src/components/Class/ClassCard/index.js
```diff
@@ -378,6 +424,7 @@ function ClassCard({ item, editClass }) {
               <ClassJoinTimerButton
                 startTime={item?.start_time}
                 link={item?.meet_link}
+                setTimeZone={setTimeZone}
```
jschanker: With the suggested renames, we'd have:
```suggestion
                onCanJoin={() => setCanJoin(true)}
```

src/components/Class/ClassCard/index.js
```diff
@@ -48,6 +50,7 @@ function ClassCard({ item, editClass }) {
   const [indicator, setIndicator] = React.useState(true);
   const [loading, setLoading] = React.useState(false);
   const user = useSelector(({ User }) => User);
+  const [timeZone, setTimeZone] = useState(false);
```
jschanker:
```suggestion
  const [canJoin, setCanJoin] = useState(false);
```

src/components/Class/ClassCard/index.js
```diff
@@ -314,6 +335,13 @@ function ClassCard({ item, editClass }) {
               >
                 <Typography textAlign="center">Edit</Typography>
               </MenuItem>
+              {item.pathway_id === 7 && (
+                <MergeClass
+                  itemID={item.id}
+                  PathwayID={item.pathway_id}
+                  pathwayFilter={pathwayFilter}
+                />
+              )}
```
komalahire: Pathway id should not be hard coded.

src/components/Class/ClassCard/index.js
```diff
@@ -333,11 +361,24 @@ function ClassCard({ item, editClass }) {
...
+            <img
+              className={classes.icons}
+              src={require("../assets/mergeClass.png")}
+              height="26px"
+              width="26px"
+            />
+
+            {item?.merge_class}
+          </Typography>
+        )}
```
komalahire: Pathway id should not be hard coded. here also

src/components/Class/ClassCard/index.js
```diff
@@ -368,7 +409,13 @@ function ClassCard({ item, editClass }) {
...
           {languageMap[item.lang]}
         </Typography>
-        <CardActions style={{ padding: "0px" }}>
+        <CardActions
+          sx={{
+            padding: "0px",
+            alignItems: "right",
+            justifyContent: "flex-end",
+          }}
+        >
```
komalahire: Add this in CSS file.

komalahire (changes requested):

src/components/Class/ClassList/index.js
```diff
@@ -83,94 +95,159 @@ function ClassList({ editClass, isShow }) {
...
-                item.title.toLowerCase().indexOf(e.target.value.toLowerCase()) >
-                -1
-            );
-            set_recurring_classes_data_set(filtered_recurring_classes);
-          } else {
-            set_recurring_classes_data_set(null);
-          }
-        }}
-      />
+              set_recurring_classes_data_set(filtered_recurring_classes);
```
jschanker: Use `const` instead of `let`. Also, you can define a function with this code, so you don't need to repeat it as you do below on lines 149 - 155.

jschanker (commented):
Some initial comments

src/pages/CreateClass/index.js
```diff
@@ -192,7 +188,6 @@ function ToggleClassFormModal() {
         {canSpecifyFacilitator && (
           <Grid item md={3}>
             <DrawerLeft
-              pathwayID={pathwayID}
               setPathwayId={setPathwayId}
               setPathwayName={setPathwayName}
               setNewPathways={setNewPathways}
```
Poonam-Singh-Bagh: `setNewPathways` is not being used in `DrawerLeft` component so you can remove it from here.

src/components/Class/ClassCard/index.js
Poonam-Singh-Bagh: @saloni-080601 you haven't done this yet. Please do it and resolve the conversation once you done.

PR #1043, opened 2023-08-16
scdule timing data

saloni-080601:
1. On class form when keep the timing radio button is false then it will show day wise start time and end time.
2. if keep the timing radio button is true, then time will remain the same for every day. 
3. On schedule object, we have to add one more object under schedule object on that, key will be days and value will be start time and end time, whatever we chose days it will add on schedule.

### DIFF
--- a/src/components/Class/ClassForm.js
+++ b/src/components/Class/ClassForm.js
@@ -408,12 +408,9 @@ function ClassForm({
   }, [partnerData]);
 
   const convertToIST = (d) => {
-    const b = String(d).split(/\D+/);
-    const dateInObj = new Date(
-      Date.UTC(b[0], --b[1], b[2], b[3], b[4], b[5], b[6])
-    );
+    const dateInObj = new Date(d);
     const utc = dateInObj.getTime() + dateInObj.getTimezoneOffset() * 60000;
-    return new Date(utc + 3600000 * +5.5).toISOString();
+    return new Date(utc + 3600000 * 5.5).toUTCString();
   };
   const createClass = (payload) => {
     setLoading(true);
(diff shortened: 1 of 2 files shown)


src/components/Class/ClassForm.js
```diff
@@ -267,7 +272,10 @@ function ClassForm({
       if (
         classFields.title !== "" &&
         classFields.partner_id?.length > 0 &&
-        classFields.on_days?.length > 0
+        classFields.on_days?.length > 0 &&
+        (checked
+          ? Object.keys(classFields.schedule).length === 0
+          : Object.keys(classFields.schedule).length > 1)
```
Poonam-Singh-Bagh, line 277: Why are you giving 2 different conditions when you want to do the same in both conditions?

src/components/Class/ClassForm.js
```diff
@@ -434,8 +453,10 @@ function ClassForm({
 
   const editClass = (payload) => {
     setLoading(true);
-    payload.start_time = convertToIST(payload.start_time);
-    payload.end_time = convertToIST(payload.end_time);
+    if (checked) {
+      payload = classFields.start_time;
+      payload = classFields.end_time;
+    }
```
Poonam-Singh-Bagh: It should be `payload.start_time = classFields.start_time;` and `payload.end_time = classFields.end_time;`

src/components/Class/ClassForm.js
```diff
@@ -614,13 +609,87 @@ function ClassForm({
...
+      //combining time and date
+      const classStartTime = moment(
+        `${classFields.date} ${classFields.start_time}`
+      );
+      const classEndTime = moment(
+        `${classFields.date} ${classFields.end_time}`
+      );
+
+      if (classStartTime.valueOf() >= classEndTime.valueOf()) {
+      }
```
Poonam-Singh-Bagh: What will happen when this condition gets true?

Poonam-Singh-Bagh (changes requested):

